PR #12, opened 2021-01-22
SLS-919 Add more configuration options 

AyyLam:
### What does this PR do?

This PR adds additional configurations to our Datadog CDK Construct that we already have for our [serverless plugin](https://github.com/DataDog/serverless-plugin-datadog#more-configuration-options).  

In addition, unit tests were written along with an updated README.md .

### Motivation

https://datadoghq.atlassian.net/browse/SLS-919

### Testing Guidelines

- Wrote and passed all unit tests
- Tested and deployed sample application to AWS sandbox account using `cdk deploy` and `cdk diff` with different configurations

### Additional Notes

Some configurations were not included due to how users will interact with the Datadog CDK construct and AWS CDK as a whole and provide more flexibility for the user as a result. 

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


lib/datadog.ts
```diff
@@ -8,36 +8,88 @@
...
+  props: DatadogProps
   constructor(scope: cdk.Construct, id: string, props: DatadogProps) {
     super(scope, id);
-    if (props.addLayers === undefined) {
-      props.addLayers = true;
+    this.scope = scope;
+    this.props = props;
+  }
+
+  public addLambdaFunctions(lambdaFunctions: lambda.Function[]){
```
agocs, line 42: I'm not sure what `addLambdaFunctions` is supposed to do here. It seems to do everything that the constructor should do (Setting props, choosing default values, checking for configuration errors), and then it calls `applyLayers` which seems to add the lambda functions. It might make sense to break a lot of this logic out of `addLambdaFunctions`

lib/datadog.ts
```diff
@@ -8,36 +8,88 @@
...
+    if ((this.props.enableDDTracing === true || this.props.enableDDTracing === undefined) && this.props.forwarderARN === undefined) {
+      throw new Error('A forwarderARN of the Datadog forwarder lambda function is required for Datadog Tracing (enabled by default). This can be disabled by setting enableDDTracing: false.')
+    }
+    if (this.props.injectLogContext === undefined) {
+      this.props.injectLogContext = defaultEnvVar.injectLogContext;
     }
-    if (props != undefined && props.lambdaFunctions.length > 0) {
-      const region = `${props.lambdaFunctions[0].env.region}`;
+    if (this.props != undefined && lambdaFunctions.length > 0) {
+      const region = `${lambdaFunctions[0].env.region}`;
```
agocs, line 72: Are all of the functions guaranteed to be in the same region? I'm 99% sure adding layers will break if they aren't. Do we want to check for that?

lib/datadog.ts
```diff
@@ -8,36 +8,88 @@
...
+    if (this.props.flushMetricsToLogs === undefined) {
+      this.props.flushMetricsToLogs = defaultEnvVar.flushMetricsToLogs;
+    }
+    if (this.props.flushMetricsToLogs === false && this.props.site === undefined) {
+      throw new Error('A site is required if flushMetricsToLogs is set to false.');
+    }
+    if (this.props.site === undefined) {
+      this.props.site = defaultEnvVar.site;
+    }
+    if (!(site_list.includes(this.props.site))) {
```
agocs, line 55: You might want to `toLower` `this.props.site` as a convenience thing for future maintainers/users.

README.md
```diff
@@ -38,27 +46,48 @@ class CdkStack extends cdk.Stack {
...
-[1]: https://github.com/DataDog/datadog-lambda-layer-python
-[2]: https://github.com/DataDog/datadog-lambda-layer-js
-[3]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html
-[4]: https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html
-[5]: https://github.com/DataDog/datadog-lambda-python/releases
-[6]: https://github.com/DataDog/datadog-lambda-js/releases
-[7]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-subscriptionfilter.html
+### Tracing
+
+Enable X-Ray Tracing on your Lambda functions. For more information, see [CDK documentation][9].
```
DarcyRaynerDD, line 70: I think we need more details for customer's about if they should use x-ray or not. Maybe @Hesperide could double check the copy here.

lib/datadog.ts
```diff
@@ -8,28 +8,45 @@
...
+  public addLambdaFunctions(lambdaFunctions: lambda.Function[]){
     if (this.props.addLayers === undefined) {
-      this.props.addLayers = true;
+      this.props.addLayers = defaultEnvVar.addLayers;
+    }
+    if (this.props.enableDDTracing === undefined) {
+      this.props.enableDDTracing = defaultEnvVar.enableDDTracing;
+    }
+    if ((this.props.enableDDTracing === true || this.props.enableDDTracing === undefined) && this.props.forwarderARN === undefined) {
+      throw new Error('A forwarderARN of the Datadog forwarder lambda function is required for Datadog Tracing (enabled by default). This can be disabled by setting enableDDTracing: false.')
```
DarcyRaynerDD, line 46: There are multiple ways to set the forwarderARN, not just through this plugin. We should keep it optional here and not throw an exception.

lib/datadog.ts
```diff
@@ -8,28 +8,45 @@
...
+  apiKMSKey?: string;
+  enableDDTracing?: boolean;
+  injectLogContext?: boolean;
 }
 
 export class Datadog extends cdk.Construct {
-  /** allows accessing the counter function */
   scope: cdk.Construct;
   props: DatadogProps;
+  xport: Transport;
```
DarcyRaynerDD, line 30: Unless this is a CDK convention, I think we should rename this transport. xport made me think it was a typo of export.

lib/env.ts
```diff
@@ -0,0 +1,30 @@
...
+  injectLogContext: true,
+};
+
+export function applyEnvVariables(
+    lambdas: lambda.Function[],
+    enableDDTracing: boolean,
+    injectLogContext: boolean,
+) {
+  lambdas.forEach((l) => {
+    l.addEnvironment(enableDDTracingEnvVar, enableDDTracing.toString().toLowerCase());
```
DarcyRaynerDD, line 27: This is fine for a small piece of code, but for larger code we generally prefer full length variable names, eg 'lambda'.

lib/transport.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    if (site === undefined) {
+      this.site = transportDefaults.site;
+    } else {
+      this.site = site;
+    }
+
+    if (!site_list.includes(this.site.toLowerCase())) {
+      throw new Error(
```
DarcyRaynerDD, line 54: This means we can't test the cdk with out staging url. Could we just print a warning instead?

test/env.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        Variables: {
+          [DD_HANDLER_ENV_VAR]: "hello.handler",
+          [logForwardingEnvVar]: "true",
+          [enableDDTracingEnvVar]: "true",
+          [injectLogContextEnvVar]: "true",
+        },
+      },
+    });
+  });
+  it("throws error if Datadog Tracing is enabled but forwarder is not defined", () => {
```
DarcyRaynerDD, line 99: Again, we probably don't need this behavior.

test/env.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      Environment: {
+        Variables: {
+          [DD_HANDLER_ENV_VAR]: "hello.handler",
+          [logForwardingEnvVar]: "true",
+          [enableDDTracingEnvVar]: "true",
+          [injectLogContextEnvVar]: "true",
+        },
+      },
+    });
+  });
```
DarcyRaynerDD, line 175: These test look good

DarcyRaynerDD (changes requested):
Mostly LGTM. Most of the comments are nit picks, but we should change the mandatory forwarderARN behaviour.

lib/datadog.ts
AyyLam: Gotcha, removed this logic from the file.

lib/datadog.ts
AyyLam: Agreed. I've changed it to `transport: Transport`

lib/env.ts
AyyLam: I changed this from `l` to `lam` as there is a naming conflict if we used `lambda`. The lambda CDK library is imported under the `lambda` alias and is best practice. Let me know if this works for you or if you have another suggestion.

lib/transport.ts
AyyLam: Gotcha, fixed it by console.logging out the error message instead. 

```
if (!site_list.includes(this.site.toLowerCase())) {
      console.log("Warning: Invalid site URL. Must be either datadoghq.com, datadoghq.eu, us3.datadoghq.com, or ddog-gov.com.")
    }
```

test/env.spec.ts
AyyLam: Removed the test associated with it 👍

test/env.spec.ts
AyyLam: Thanks! 😃

DarcyRaynerDD (approved):

PR #18, opened 2021-02-12
dd-cdk CI/CD workflows build & CDK peer dependencies

zARODz11z:
### What does this PR do?
This PR adds a build Github workflow that should be ran every time a push is made to any branch on the dd-cdk repository. This PR also fixes a type error in the unit tests by casting the handler name to a string.

### Motivation
I had previously made a PR where I thought I addressed the peer dependency issue, after I merged my branch to main and attempted to make a new release I found out some of the unit tests failed, having this GitHub workflow will hopefully mitigate a faulty merge like mine in the future.

### Additional Notes
This build workflow was ported over from https://github.com/DataDog/serverless-plugin-datadog/blob/master/.github/workflows/build.yml

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "datadog-cdk-constructs",
-  "version": "0.1.6",
+  "version": "0.1.7",
```
nhinsch, line 3: You don't need to increment the version number here unless you are actually publishing the changes to npm as a new release. Actually, I think we would ideally change the version number in the `package.json` as part of the release script.

.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Node 12.7
```
nhinsch, line 13: For the Lint job, let's run it in the current LTS (long term support) version of Node, which is `14`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Lint
+        run: yarn lint
+  
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        node-version: [10.17, 12.7]
```
nhinsch, line 40: Let's test Node version 14 here as well... we can do `10`, `12`, and `14`. If we leave out the minor version number, I believe we will just use the latest minor version within that major version, which is probably what we want here.

.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
...
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
+        run: yarn install
+
+      - name: Build
+        run: yarn build
+
+      - name: Run tests
+        run: yarn test
+
+      - name: Upload code coverage report
```
nhinsch, line 68: Has this been set up? I don't know a lot about how the `codecov` tool works. But it's worth checking to see if this is working correctly, or maybe just remove this for now.

package.json
```diff
@@ -31,9 +30,15 @@
     "typescript": "^3.9.7"
   },
   "dependencies": {
```
nhinsch, line 32: I thought we did not want these as dependencies any more? Just peer dependencies?

zARODz11z: datadog.ts imports from aws-cdk/core and the other peer dependencies. When I would remove the regular dependecies from the package.json I would get a ton of import errors from the internal files that used any of those cdk packages. It seems like just having it as a dev dependency does not add it to the local node modules therefore the dd-cdk files cannot properly import.

nhinsch (approved):

PR #20, opened 2021-02-18
Duplicate Constructs Bug 🐛

zARODz11z:
### What does this PR do?
This PR changes the logic for the creation of the `forwarder`, `subscriptionFilter`, and `layer` constructs in an attempt to never create duplicate constructs in the same cdk construct tree while also methodically creating new constructs only when necessary.

A customer ran into this error `There is already a Construct with name 'forwarder' in ...`. After doing some research I was able to reproduce the error by creating a unit test that created a mock aws-cdk app, stack, and constructs related to the Datadog-cdk-construct library. This unit test intentionally called the `addForwarder` function twice. I am not sure if this is exactly how the customer encountered this problem but I believe my solution should prevent it from happening again.
Before both `addForwarder` calls our construct tree looked like this:  
app/
|─ stack/
│  ├─ AssetParameters
│  ├─ NodeHandler

After 1 `addForwarder` call the construct tree looked like this (as you can see the forwarder construct was successfully added):
app/
|─ stack/
│  ├─ AssetParameters
│  ├─ forwarder
│  ├─ LogRetention
│  ├─ NodeHandler

Now the problem occurs when the `addForwarder` function is called for the second time. Since the forwarder construct already exists in the construct tree the deployment will fail here because there cannot be duplicate constructs in the same cdk construct tree. As illustrated with the picture below, the 2nd call is trying to place another `forwarder` construct in the tree.
app/
|─ stack/
│  ├─ AssetParameters
│  ├─ forwarder
│  ├─ forwarder //this causes the error
│  ├─ LogRetention
│  ├─ NodeHandler

### Testing Guidelines
I created a unit tests in the index.spec.ts file to test my changes while also building and packing my local version of the dd-cdk-constructs and testing the unit tests manually in my [lambda-cdk-constructs-demo project](https://github.com/DataDog/cloud-integrations-lambdas/tree/master/lambda-cdk-constructs-demo).

### Additional Notes
This is the error I produced compared to the error on the [ticket](https://datadoghq.atlassian.net/jira/software/projects/SLS/boards/205?selectedIssue=SLS-973) 
addForwarder  does not create duplicate forwarder constructs

I learned about the construct tree here https://docs.aws.amazon.com/cdk/latest/guide/constructs.html, also this picture helped a lot.
Construct
https://d2908q01vomqb2.cloudfront.net/da4b9237bacccdf19c0760cab7aec4a8359010b0/2018/12/17/appstack.png

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR passes the integration tests (ask a Datadog member to run the tests)

### DIFF
--- a/lib/forwarder.ts
+++ b/lib/forwarder.ts
@@ -13,6 +13,9 @@ import * as crypto from "crypto";
 import { LambdaDestination } from "@aws-cdk/aws-logs-destinations";
 const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
 export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderARN: string) {
+  if (scope.node.tryFindChild("forwarder")) {
+    return;
+  }
   const forwarder = lambda.Function.fromFunctionArn(scope, "forwarder", forwarderARN);
   const forwarderDestination = new LambdaDestination(forwarder);
   lambdaFunctions.forEach((lam) => {
--- a/lib/forwarder.ts
+++ b/lib/forwarder.ts
@@ -12,25 +12,35 @@ import { FilterPattern } from "@aws-cdk/aws-logs";
 import * as crypto from "crypto";
 import { LambdaDestination } from "@aws-cdk/aws-logs-destinations";
 const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
-export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderARN: string) {
-  const forwarderConstructID = "forwarder" + crypto.createHash("sha256").update(forwarderARN).digest("hex");
+
+function generateForwaderConstrucId(forwarderArn: string) {
+  return "forwarder" + crypto.createHash("sha256").update(forwarderArn).digest("hex");
+}
+function generateSubscriptionFilterName(functionArn: string, forwarderArn: string) {
+  const subscriptionFilterValue: string = crypto
+    .createHash("sha256")
+    .update(functionArn)
+    .update(forwarderArn)
+    .digest("hex");
+  const subscriptionFilterValueLength = subscriptionFilterValue.length;
+  const subscriptionFilterName =
+    SUBSCRIPTION_FILTER_PREFIX +
+    subscriptionFilterValue.substring(subscriptionFilterValueLength - 8, subscriptionFilterValueLength);
+
+  return subscriptionFilterName;
+}
+
+export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderArn: string) {
+  const forwarderConstructID = generateForwaderConstrucId(forwarderArn);
   let forwarder;
   if (scope.node.tryFindChild(forwarderConstructID)) {
     forwarder = scope.node.tryFindChild(forwarderConstructID) as lambda.IFunction;
   } else {
-    forwarder = lambda.Function.fromFunctionArn(scope, forwarderConstructID, forwarderARN);
+    forwarder = lambda.Function.fromFunctionArn(scope, forwarderConstructID, forwarderArn);
   }
   const forwarderDestination = new LambdaDestination(forwarder);
   lambdaFunctions.forEach((lam) => {
-    const subscriptionFilterValue: string = crypto
-      .createHash("sha256")
-      .update(lam.functionArn)
-      .update(forwarderARN)
-      .digest("hex");
-    const subscriptionFilterValueLength = subscriptionFilterValue.length;
-    const subscriptionFilterName =
-      SUBSCRIPTION_FILTER_PREFIX +
-      subscriptionFilterValue.substring(subscriptionFilterValueLength - 8, subscriptionFilterValueLength);
+    const subscriptionFilterName = generateSubscriptionFilterName(lam.functionArn, forwarderArn);
     lam.logGroup.addSubscriptionFilter(subscriptionFilterName, {
       destination: forwarderDestination,
       filterPattern: FilterPattern.allEvents(),
--- a/lib/forwarder.ts
+++ b/lib/forwarder.ts
@@ -13,21 +13,27 @@ import * as crypto from "crypto";
 import { LambdaDestination } from "@aws-cdk/aws-logs-destinations";
 const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
 export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderARN: string) {
-  if (scope.node.tryFindChild("forwarder")) {
-    return;
+  const forwarderConstructID = "forwarder" + crypto.createHash("sha256").update(forwarderARN).digest("hex");
+  let forwarder;
+  if (scope.node.tryFindChild(forwarderConstructID)) {
+    forwarder = scope.node.tryFindChild(forwarderConstructID) as lambda.IFunction;
+  } else {
+    forwarder = lambda.Function.fromFunctionArn(scope, forwarderConstructID, forwarderARN);
   }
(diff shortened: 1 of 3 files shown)


lib/forwarder.ts
```diff
@@ -13,6 +13,9 @@ import * as crypto from "crypto";
 import { LambdaDestination } from "@aws-cdk/aws-logs-destinations";
 const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
 export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderARN: string) {
+  if (scope.node.tryFindChild("forwarder")) {
```
tianchu, line 16: This means the rest of this function's logic (subscribing the forwarder to log groups) won't execute on the second call, do you know if the function is called with the same `lambdaFunctions` and `forwarderARN` on the subsequent calls? in other words, do you know if it's safe to NOT create the subscription filters for the subsequent calls because they are already created?

test/index.spec.ts
```diff
@@ -1,12 +1,119 @@
...
+      handler: "hello.handler",
+    });
+    addForwarder(stack, [nodeLambda], "forwarder-arn");
+    addForwarder(stack, [pythonLambda], "forwarder-arn");
+    const nodeLambdaLogGroupSubscriptionFilterName = createSubscriptionFilterName(nodeLambda.functionArn,"forwarder-arn")
+    const pythonLambdaLogGroupSubscriptionFilterName = createSubscriptionFilterName(pythonLambda.functionArn,"forwarder-arn")
+    expect(nodeLambda.logGroup.node.tryFindChild(nodeLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
+    expect(pythonLambda.logGroup.node.tryFindChild(pythonLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
+  });
+  it("Subscribes two different forwarders to two different lambda functions via seperate addForwarder function calls",() => {
```
nhinsch, line 69: Do we currently support multiple forwarders?

test/index.spec.ts
```diff
@@ -1,12 +1,119 @@
...
+      handler: "hello.handler",
+    });
+    addForwarder(stack, [nodeLambda], "forwarder-arn");
+    addForwarder(stack, [pythonLambda], "forwarder-arn2");
+    const nodeLambdaLogGroupSubscriptionFilterName = createSubscriptionFilterName(nodeLambda.functionArn,"forwarder-arn")
+    const pythonLambdaLogGroupSubscriptionFilterName = createSubscriptionFilterName(pythonLambda.functionArn,"forwarder-arn2")
+    expect(nodeLambda.logGroup.node.tryFindChild(nodeLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
+    expect(pythonLambda.logGroup.node.tryFindChild(pythonLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
+  });
+  it("Subscribes the same forwarder to two different lambda functions via one addForwarder function call",() => {
```
nhinsch, line 93: Formatting nit: I would put a blank line between each test, it makes things a bit easier to read.

lib/forwarder.ts
```diff
@@ -13,18 +13,27 @@ import * as crypto from "crypto";
...
+  let forwarder;
+  if (scope.node.tryFindChild(forwarderConstructID)) {
+    forwarder = scope.node.tryFindChild(forwarderConstructID) as lambda.IFunction;
+  } else {
+    forwarder = lambda.Function.fromFunctionArn(scope, forwarderConstructID, forwarderARN);
+  }
   const forwarderDestination = new LambdaDestination(forwarder);
   lambdaFunctions.forEach((lam) => {
-    const subscriptionFilterValue: string = crypto.createHash("sha256").update(lam.functionArn).digest("hex");
+    const subscriptionFilterValue: string = crypto
```
nhinsch, line 25: I would suggest pulling this logic out into a separate helper function, maybe `generateSubscriptionFilterName(functionArn, forwarderArn)`. Then the logic for `addForwarder()` itself becomes simpler.

lib/forwarder.ts
```diff
@@ -13,18 +13,27 @@ import * as crypto from "crypto";
 import { LambdaDestination } from "@aws-cdk/aws-logs-destinations";
 const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
 export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderARN: string) {
```
nhinsch, line 15: Nit: I know this was already there, but for abbreviations in camel case like ARN, they should be capitalized like `Arn` instead of `ARN`. Also this is more consistent with other usages of `Arn` in the same file.

nhinsch: I actually don't think it matters much which way it is capitalized, but I would at least try to make it consistent within the codebase ideally. Otherwise it could get a bit confusing.

lib/forwarder.ts
nhinsch: Two smaller functions is pretty much always better than one big function with lots of details in my book.

test/index.spec.ts
```diff
@@ -1,12 +1,119 @@
...
 } from "../lib/redirect";
+const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
+
+function createSubscriptionFilterName(lambdaFunctionArn:string,forwarderArn:string){
+  const subscriptionFilterValue: string = crypto.createHash("sha256").update(lambdaFunctionArn).update(forwarderArn).digest("hex");
+    const subscriptionFilterValueLength = subscriptionFilterValue.length;
+    const subscriptionFilterName = SUBSCRIPTION_FILTER_PREFIX + subscriptionFilterValue.substring(subscriptionFilterValueLength - 8, subscriptionFilterValueLength);
+    return subscriptionFilterName;
+}
+describe("addForwarder", () => {
```
nhinsch, line 20: These tests are great! 🥇

test/index.spec.ts
nhinsch: (discussed on Slack -- it is a possibility if the construct is used twice)

lib/layer.ts
```diff
@@ -105,3 +105,8 @@ export function getMissingLayerVersionErrorMsg(functionKey: string, formalRuntim
     `Please add the '${paramRuntime}LayerVersion' parameter for the Datadog serverless macro.`
   );
 }
+
+function generateLayerID(lambdaFunctionArn: string, runtime: string) {
+  const layerValue: string = crypto.createHash("sha256").update(lambdaFunctionArn).digest("hex");
+  return layerPrefix + "-" + runtime + "-" + layerValue;
```
nhinsch, line 111: What is `layerValue` here? It looks like you removed that variable?

zARODz11z: layerValue used to be a number that would start at 0, be appended to the layerPrefix on the old line 79, and simply be incremented each time a layer was added for each lambda function. This turned out to be a problem because I was getting duplicate layer constructs due to the simple naming convention. To fix this I deleted the layerValue variable in applyLayers and delegated creating the unique Id in the generateLayerId function where the new layerValue was a hash of the lambda functionArn. Ultimately the layer Id was the layer prefix (DatadogLayer) appended to the runtime and the hash value created on line 110.

nhinsch: Ah got it, makes sense

nhinsch (approved):
This looks good to me

zARODz11z:
> This looks good to me

For the record I manually tested these scenarios and achieved to expected result:
1) Subscribes the same forwarder to two different lambda functions via separate addLambdaFunctions function calls
2) Subscribes the same forwarder to two different lambda functions via one addForwarder function call
3) Throws an error when a customer redundantly calls the addLambdaFunctions function on the same lambda function(s) and forwarder
4) Subscribes two different forwarders to two different lambda functions via separate addForwarder function calls

PR #21, opened 2021-02-23
Datadog CDK Constructs Extension Layer Support

zARODz11z:
### What does this PR do?
This PR adds logic to support the Datadog Extension Layer in the Datadog CDK Construct library.

### Motivation
The extension layer is gaining more traction and may provide a easier alternative then using the Datadog forwarder process we have in place now. The customer should have the choice between using the extension layer or the forwarder process.

### Testing Guidelines
I created new unit tests covering the scenarios I could come up with while also linking the development copy of the DD-CDK to my [CDK-Demo](https://github.com/DataDog/cloud-integrations-lambdas/tree/master/lambda-cdk-constructs-demo) application where I manually tested each of the scenarios while confirming expected results through the AWS console and Datadog app.

### Additional Notes
This branch was created off of one of my previous in progress tickets so I had to merge main into this branch while also resolving any merge requests therefore thats why the commit history is strange. Also I dedicated the addLayer function to create and add the layer to each lambda function in efforts of minimizing duplicate code when adding the extension layer. 

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


lib/layer.ts
```diff
@@ -99,14 +122,27 @@ export function getLayerARN(region: string, version: number, runtime: string) {
...
   );
 }
 
-function generateLayerId(lambdaFunctionArn: string, runtime: string) {
+function generateLambdaLayerId(lambdaFunctionArn: string, runtime: string) {
   const layerValue: string = crypto.createHash("sha256").update(lambdaFunctionArn).digest("hex");
   return layerPrefix + "-" + runtime + "-" + layerValue;
 }
+
+function generateExtensionLayerId(lambdaFunctionArn: string, runtime: string) {
```
nhinsch, line 145: Do we really have a concept of runtime for the extension layer? I.e. for the normal Lambda layer, there is a difference between the Node layer and the Python layer (different layers for different runtimes). But for the extension, the same layer works for all runtimes.

test/layer.spec.ts
```diff
@@ -153,20 +213,79 @@ describe("isGovCloud", () => {
...
+    datadogCdk.addLambdaFunctions([hello]);
+    expect(stack).toHaveResource("AWS::Lambda::Function", {
+      Layers: [
+        `arn:aws-us-gov:lambda:us-gov-east-1:${DD_GOV_ACCOUNT_ID}:layer:Datadog-Node10-x:${NODE_LAYER_VERSION}`,
+        `arn:aws-us-gov:lambda:us-gov-east-1:${DD_GOV_ACCOUNT_ID}:layer:Datadog-Extension:${EXTENSION_LAYER_VERSION}`,
+      ],
+    });
+  });
+
+  it("adds an extension layer along with a python layer ", () => {
```
nhinsch, line 262: Nit: I would update the description of this test to make clear it pertains to GovCloud

nhinsch:
I believe the desired behavior is that when an extension layer is added, we should **not** subscribe functions to the Forwarder. Maybe we should throw an error if the customer tries to add an extension layer but also a Forwarder ARN.

Also, customers should be required to set the `apiKey` (or `apiKmsKey`) and the `site` when the extension layer is used, because without that, the extension won't know where to send logs.

I would add these behaviors and also make sure the README is up to date explaining them.

lib/layer.ts
```diff
@@ -99,14 +122,27 @@ export function getLayerARN(region: string, version: number, runtime: string) {
...
+
 export function getMissingLayerVersionErrorMsg(functionKey: string, formalRuntime: string, paramRuntime: string) {
   return (
     `Resource ${functionKey} has a ${formalRuntime} runtime, but no ${formalRuntime} Lambda Library version was provided. ` +
     `Please add the '${paramRuntime}LayerVersion' parameter for the Datadog serverless macro.`
   );
 }
 
-function generateLayerId(lambdaFunctionArn: string, runtime: string) {
+function generateLambdaLayerId(lambdaFunctionArn: string, runtime: string) {
```
nhinsch, line 140: If we don't already have them, it might be nice to add some quick tests for `generateLambdaLayerId` and `generateExtensionLayerId`

README.md
```diff
@@ -58,6 +58,7 @@ To further configure your Datadog construct, use the following custom parameters
 | `addLayers`          | Whether to add the Lambda Layers or expect the user to bring their own. Defaults to true. When true, the Lambda Library version variables are also required. When false, you must include the Datadog Lambda library in your functions' deployment packages.         |
 | `pythonLayerVersion` | Version of the Python Lambda layer to install, such as 21. Required if you are deploying at least one Lambda function written in Python and `addLayers` is true. Find the latest version number from [https://github.com/DataDog/datadog-lambda-python/releases][5]. |
 | `nodeLayerVersion`   | Version of the Node.js Lambda layer to install, such as 29. Required if you are deploying at least one Lambda function written in Node.js and `addLayers` is true. Find the latest version number from [https://github.com/DataDog/datadog-lambda-js/releases][6].   |
+| `extensionLayerVersion`   | Version of the Datadog Lambda Extension layer to install, such as 5. The Datadog Lambda Extension layer is in public preview. Learn more about the lambda extension here [https://docs.datadoghq.com/serverless/datadog_lambda_library/extension][12].   |
```
nhinsch, line 61: Nit: I think you want to make it so that the word `here` is the link

lib/layer.ts
zARODz11z: Removed that parameter, will be included in the next commit.

test/layer.spec.ts
zARODz11z: Change included in next commit

README.md
zARODz11z: Modified, will be present in next commit

lib/layer.ts
zARODz11z: I tried to create unit tests for the generateLambdaLayerId, this was my approach, I create a sample lambda function, pass its functionArn and runtime to the generateLambdaLayerId function and since the Id is "-" delimited I can simply split and assert each part. However I cannot assert the part that was a hash of the function arn because the function arn is a token that is not available until synthesis of the cloud formation template. https://docs.aws.amazon.com/cdk/latest/guide/tokens.html 
Screen Shot 2021-02-23 at 3 03 16 PM

nhinsch: I would just pass in a fake Arn as a string.

zARODz11z: Tests will be included in next commit

zARODz11z:
> I believe the desired behavior is that when an extension layer is added, we should **not** subscribe functions to the Forwarder. Maybe we should throw an error if the customer tries to add an extension layer but also a Forwarder ARN.
> 
> Also, customers should be required to set the `apiKey` (or `apiKmsKey`) and the `site` when the extension layer is used, because without that, the extension won't know where to send anything.
> 
> I would add these behaviors and also make sure the README is up to date explaining them.

The site has a default of "datadoghq.com" if left undefined. If a customer wants their extension to send data to Datadoghq.com then I think it should be fine to leave "site" undefined and let the default take over. If they wish to send to a different site then I believe thats where they can set site to whatever they want. Therefore I did not create logic that required a "site" in the most recent commit. The other suggestions in this comment ^ were added at the end of the Datadog constructor in datadog.ts.

zARODz11z:
> > I believe the desired behavior is that when an extension layer is added, we should **not** subscribe functions to the Forwarder. Maybe we should throw an error if the customer tries to add an extension layer but also a Forwarder ARN.
> > Also, customers should be required to set the `apiKey` (or `apiKmsKey`) and the `site` when the extension layer is used, because without that, the extension won't know where to send anything.
> > I would add these behaviors and also make sure the README is up to date explaining them.
> 
> The site has a default of "datadoghq.com" if left undefined. If a customer wants their extension to send data to Datadoghq.com then I think it should be fine to leave "site" undefined and let the default take over. If they wish to send to a different site then I believe thats where they can set site to whatever they want. Therefore I did not create logic that required a "site" in the most recent commit. The other suggestions in this comment ^ were added at the end of the Datadog constructor in datadog.ts.

On a second thought after reading the configuration [here](https://github.com/DataDog/datadog-cdk-constructs) I believe that site is needed when the extensionLayerVersion is provided because providing the extensionLayerVersion automatically overrides flushMetricsToLogs to false so that the api key will be read and correctly added as an environment variable per our slack convo.

lib/datadog.ts
```diff
@@ -32,11 +33,25 @@ export class Datadog extends cdk.Construct {
...
+        throw new Error("`extensionLayerVersion` and `forwarderArn` cannot be set at the same time.");
+      }
+      if (this.props.apiKey === undefined && this.props.apiKMSKey === undefined) {
+        throw new Error("When `extensionLayer` is set, `apiKey` or `apiKMSKey` must also be set.");
+      }
+      if (this.props.site === undefined) {
+        throw new Error("When `extensionLayer` is set, `site` must also be set.");
+      }
+    }
+
```
zARODz11z, line 48: I figured this ^ would be the easiest place to check the props and throw the errors accordingly. https://github.com/DataDog/datadog-cdk-constructs/pull/21#issuecomment-784497365

lib/layer.ts
```diff
@@ -99,14 +122,27 @@ export function getLayerARN(region: string, version: number, runtime: string) {
...
   );
 }
 
-function generateLayerId(lambdaFunctionArn: string, runtime: string) {
+export function generateLambdaLayerId(lambdaFunctionArn: string, runtime: string) {
   const layerValue: string = crypto.createHash("sha256").update(lambdaFunctionArn).digest("hex");
   return layerPrefix + "-" + runtime + "-" + layerValue;
 }
+
+export function generateExtensionLayerId(lambdaFunctionArn: string) {
```
zARODz11z, line 145: removed runtime here to address https://github.com/DataDog/datadog-cdk-constructs/pull/21#discussion_r581366356

lib/transport.ts
```diff
@@ -26,14 +26,25 @@ export class Transport {
...
+  ) {
     if (flushMetricsToLogs === undefined) {
       this.flushMetricsToLogs = transportDefaults.flushMetricsToLogs;
     } else {
       this.flushMetricsToLogs = flushMetricsToLogs;
     }
 
+    if (extensionLayerVersion !== undefined) {
+      this.flushMetricsToLogs = false;
+    }
```
zARODz11z, line 46: If the customer requests to install the extension layer ^ then we override flushMetricsToLogs to be false so that the apiKey will be read and added as an environment variable. Addresses our slack convo

test/layer.spec.ts
```diff
@@ -28,17 +33,179 @@ describe("applyLayers", () => {
...
+    }catch(e){
+      threwError = true;
+      thrownError = e;
+    }
+    expect(threwError).toBe(true);
+    expect(thrownError?.message).toEqual("`extensionLayerVersion` and `forwarderArn` cannot be set at the same time.");
+
+  });
+
+  it("throws an error when the `extensionLayerVersion` is set while `site` is undefined", () => {
```
zARODz11z, line 176: I am unsure if this unit test is appropriate because "site" is said to have a default value of datadoghq.com but the logic I have right now [here](https://github.com/DataDog/datadog-cdk-constructs/pull/21/files#diff-e07b5220c79ce96edc53faf651cde1714026204d4cfc4195eed15f516145fc8aR44) enforces that if the `extensionLayerVersion` is set then the customer needs to also set `site`.

test/layer.spec.ts
```diff
@@ -153,20 +320,114 @@ describe("isGovCloud", () => {
...
+      code: lambda.Code.fromAsset("test"),
+      handler: "hello.handler",
+    });
+    const lambdaFunctionArn = "functionArn";
+    const runtime: string = hello.runtime.name;
+    const lambdaLayerId: string = generateLambdaLayerId(lambdaFunctionArn,runtime);
+    const lambdaLayerIdParts: string[] = lambdaLayerId.split("-");
+    expect(lambdaLayerIdParts[0]).toEqual("DatadogLayer");
+    expect(lambdaLayerIdParts[1]).toEqual("nodejs10.x");
+    expect(crypto.createHash("sha256").update(lambdaFunctionArn).digest("hex")).toEqual(lambdaLayerIdParts[2]);
```
zARODz11z, line 421: There very well could be a better way of testing this function along with the generateExtensionLayerId function but this is what came to mind.  I am open to suggestions.

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
+| `flushMetricsToLogs` | Send custom metrics using Cloudwatch logs with the Datadog Forwarder Lambda function (recommended). Defaults to `true`. If you disable this parameter or set `extensionLayerVersion`, it's required to set the parameters `site` and `apiKey` (or `apiKMSKey` if encrypted). |
+| `site`               | Set which Datadog site to send data, only needed when flushMetricsToLogs is `false` or `extensionLayerVersion` is set. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com` and `ddog-gov.com`. The default is `datadoghq.com`. |
+| `apiKey`             | Datadog API Key, only needed when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set. For more information about getting a Datadog API key, see the [API key documentation][8]. |
+| `apiKMSKey`          | Datadog API Key encrypted using KMS. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set, and you are using KMS encryption. |
 | `enableDDTracing`    | Enable Datadog tracing on your Lambda functions. Defaults to `true`. When enabled, it's required to set the `forwarderARN` parameter. |
```
zARODz11z, line 67: I added the warnings above multiple times in each of the related parameters, I read it multiple times and could not think of a better way of writing it. Im open to suggestions on this.

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
-| `site`               | Set which Datadog site to send data, only needed when flushMetricsToLogs is `false`. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com` and `ddog-gov.com`. The default is `datadoghq.com`. |
-| `apiKey`             | Datadog API Key, only needed when `flushMetricsToLogs` is `false`. For more information about getting a Datadog API key, see the [API key documentation][8]. |
-| `apiKMSKey`          | Datadog API Key encrypted using KMS. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false`, and you are using KMS encryption. |
+| `extensionLayerVersion`   | Version of the Datadog Lambda Extension layer to install, such as 5. When `extensionLayerVersion` is set, `site` must be set, and `apiKey` (or `apiKMSKey` if encrypted) need to be set. While using `extensionLayerVersion` do not set `forwarderARN`.The Datadog Lambda Extension layer is in public preview. Learn more about the lambda extension [here][12].   |
```
nhinsch, line 61: Nits:

When `extensionLayerVersion` is set, `site` must be set, and `apiKey` (or `apiKMSKey` if encrypted) need to be set. ---> When `extensionLayerVersion` is set, `site` and `apiKey` (or `apiKMSKey`) need to be set as well.

Also: `not set `forwarderARN`.The` -> `not set `forwarderARN`. The` (missing space)

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
+| `forwarderARN`         | When set, the plugin will automatically subscribe the functions' log groups to the Datadog Forwarder. Do not set `forwarderARN` when `extensionLayerVersion` is set.|
```
nhinsch, line 62: I know this was already written, but this is backwards, the Forwarder is subscribed to the functions' log groups, not the other way around.

README.md
nhinsch: I think that makes sense, people might just look up one specific parameter, and they should see the warning for that parameter.

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
+| `flushMetricsToLogs` | Send custom metrics using Cloudwatch logs with the Datadog Forwarder Lambda function (recommended). Defaults to `true`. If you disable this parameter or set `extensionLayerVersion`, it's required to set the parameters `site` and `apiKey` (or `apiKMSKey` if encrypted). |
```
nhinsch, line 63: I think you can change `If you disable this parameter or set `extensionLayerVersion`, it's required...` to `If you disable this parameter, it's required...`

That is, I don't think you need to discuss the implications of setting `extensionLayerVersion` in the section about `flushMetricsToLogs`

lib/datadog.ts
nhinsch: This logic looks good, I would extract it into a separate function (`validateProps()` maybe?) and add tests for the validation logic. Also, do we need to validate that if flushMetricsToLogs is false, then an API key must be set?

lib/transport.ts
nhinsch: I would add a short comment in the code explaining why we set flushMetricsToLogs as false: if the extension is used, metrics will be submitted via the extension.

Also, can we just make it so that we ALWAYS set the API key on the Lambda functions if an API key is provided? There is no downside I am aware of in setting that environment variable, so even if flush metrics to logs is enabled and it's not strictly necessary, it seems that going out of our way to NOT set the API key if we have one just makes things more complicated.

lib/datadog.ts
nhinsch: Actually, I see that you already have tests for the validation (nice!!) but I still think it would be a good idea to factor this out into a smaller function. It would actually be easier to write tests for the smaller validation function. But since you already wrote tests for the overarching function, no need to re-do them. They should still work the same with this logic factored out into a smaller function.

test/layer.spec.ts
nhinsch: I think you could simplify this test a lot by just passing in strings.

`GenerateLambdaLayerId()` just takes in two strings and returns a new string. It doesn't really matter where those strings come from. So, all the stuff about creating a Stack with a Lambda function, then pulling the runtime out of it, doesn't really have anything directly to do with `GenerateLambdaLayerId()`.

I think it would be sufficient to have a really simple test here: pass in two strings and confirm that the result is what we would expect. Same for `generateExtensionLayerId`.

test/layer.spec.ts
```diff
@@ -153,20 +320,114 @@ describe("isGovCloud", () => {
...
+  });
+});
+
+describe("generateExtensionLayerId", () => {
+  it("generates an extension ID consisting of the prefix and hash value delimited by hyphens", () => {
+    const lambdaFunctionArn = "functionArn";
+    const lambdaLayerId: string = generateExtensionLayerId(lambdaFunctionArn);
+    const lambdaLayerIdParts: string[] = lambdaLayerId.split("-");
```
nhinsch, line 429: I think we can simplify this to just test that the overall result string is what we expect it to be. We don't have to necessarily pull apart the result string into pieces and check each one.

test/layer.spec.ts
nhinsch: Ah that is a good point -- I think actually maybe we DON'T want to enforce that `site` must be set when `extensionLayerVersion` is set. Because, like you point out, we have a default value. So maybe we should remove that enforcement check. Unlike the API key, for which there obviously is no default, so they really do need to pass one in.

lib/transport.ts
zARODz11z: changes in setEnvVars() in transport.ts file should address the comment above ^

test/layer.spec.ts
zARODz11z: I removed the enforcement of a site when extension layer version is provided, therefore a site is not required when the extension layer is provided however if they do provide it then it will be correctly added. I updated the readme accordingly

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
+| `extensionLayerVersion`   | Version of the Datadog Lambda Extension layer to install, such as 5. When `extensionLayerVersion` is set, `apiKey` (or `apiKMSKey`) needs to be set as well. While using `extensionLayerVersion` do not set `forwarderARN`. The Datadog Lambda Extension layer is in public preview. Learn more about the lambda extension [here][12].   |
+| `forwarderARN`         | When set, the plugin will automatically subscribe the Datadog Forwarder to the functions' log groups. Do not set `forwarderARN` when `extensionLayerVersion` is set.|
+| `flushMetricsToLogs` | Send custom metrics using Cloudwatch logs with the Datadog Forwarder Lambda function (recommended). Defaults to `true`. If you disable this parameter, it's required to set `apiKey` (or `apiKMSKey`). |
+| `site`               | Set which Datadog site to send data, when flushMetricsToLogs is `false` or `extensionLayerVersion` is set you may set any of the possible sites or just leave it to the default value. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com` and `ddog-gov.com`. The default is `datadoghq.com`. |
```
nhinsch, line 64: I would rephrase this as:

Set which Datadog site to send data. This is only used when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set. (keeping the rest the same)

lib/datadog.ts
```diff
@@ -62,4 +72,30 @@ export class Datadog extends cdk.Construct {
       this.transport.setEnvVars(lambdaFunctions);
     }
   }
+
+  public validateProps(props: DatadogProps) {
```
nhinsch, line 76: Do you think validateProps should be a public method on the construct class, or should it just be a standalone function? I would lean towards making it a standalone function, just so it's easier to test. Also, I doubt it should be public, not sure why other code would need to call this.

test/transport.spec.ts
```diff
@@ -141,7 +206,7 @@ describe("logForwardingEnvVar", () => {
 });
 
 describe("apiKeyEnvVar", () => {
```
nhinsch, line 208: Since we no longer care about whether flushMetricsToLogs is true or false here, I don't think we need both of these test cases. We just need to test that the environment variable is set.

test/transport.spec.ts
```diff
@@ -284,7 +354,8 @@ describe("apiKMSKeyEnvVar", () => {
       },
     });
   });
-  it("does not set DD_KMS_API_KEY environment variable when flushMetricsToLogs is true", () => {
+
+  it("adds DD_KMS_API_KEY environment variable when flushMetricsToLogs is true", () => {
```
nhinsch, line 358: Likewise here, I think we can just have one test case, the behavior is now the same when flushMetricsToLogs is true/false.

lib/datadog.ts
```diff
@@ -62,4 +72,30 @@ export class Datadog extends cdk.Construct {
...
+    if (props.site !== undefined && !siteList.includes(props.site.toLowerCase())) {
+      console.log(
+        "Warning: Invalid site URL. Must be either datadoghq.com, datadoghq.eu, us3.datadoghq.com, or ddog-gov.com.",
+      );
+    }
+    if (
+      (props.apiKey !== undefined && props.apiKMSKey !== undefined && props.flushMetricsToLogs === false) ||
+      (props.apiKey === undefined && props.apiKMSKey === undefined && props.flushMetricsToLogs === false)
+    ) {
+      throw new Error(
```
nhinsch, line 96: I know this was already here, but it seems like this is the wrong error message for `(props.apiKey === undefined && props.apiKMSKey === undefined && props.flushMetricsToLogs === false)`. I.e. when neither is set, why do we tell the user that they are mutually exclusive? Maybe it would be better to say: "When `flushMetricsToLogs` is false, `apiKey` or `apiKMSKey` must also be set." in that situation. Kind of like we do for when there is an extension layer version.

nhinsch: In other words, I think these two situations deserve two different error messages.

lib/datadog.ts
```diff
@@ -62,4 +72,30 @@ export class Datadog extends cdk.Construct {
...
+  public validateProps(props: DatadogProps) {
+    const siteList: string[] = ["datadoghq.com", "datadoghq.eu", "us3.datadoghq.com", "ddog-gov.com"];
+
+    if (props.extensionLayerVersion !== undefined) {
+      if (props.forwarderARN !== undefined) {
+        throw new Error("`extensionLayerVersion` and `forwarderArn` cannot be set at the same time.");
+      }
+      if (props.apiKey === undefined && props.apiKMSKey === undefined) {
+        throw new Error("When `extensionLayer` is set, `apiKey` or `apiKMSKey` must also be set.");
+      }
```
nhinsch, line 85: What happens when `extensionlayer` is set, `apiKey` is set, AND `apiKMSKey` is set?

It seems like we never want `apiKey` and `KMSKey` to both be set, regardless of all other settings. We can probably refactor this code to reflect that.

test/layer.spec.ts
nhinsch: That sounds good as long as we are defaulting to `datadoghq.com`.

lib/datadog.ts
zARODz11z: Yeah I thought the way it was phrased before was odd

lib/datadog.ts
zARODz11z: I refactored it in datadog.ts

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
-| `site`               | Set which Datadog site to send data, only needed when flushMetricsToLogs is `false`. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com` and `ddog-gov.com`. The default is `datadoghq.com`. |
-| `apiKey`             | Datadog API Key, only needed when `flushMetricsToLogs` is `false`. For more information about getting a Datadog API key, see the [API key documentation][8]. |
-| `apiKMSKey`          | Datadog API Key encrypted using KMS. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false`, and you are using KMS encryption. |
+| `extensionLayerVersion`   | Version of the Datadog Lambda Extension layer to install, such as 5. When `extensionLayerVersion` is set, `apiKey` (or `apiKMSKey`) needs to be set as well. While using `extensionLayerVersion` do not set `forwarderARN`. The Datadog Lambda Extension layer is in public preview. Learn more about the lambda extension [here][12].   |
```
nhinsch, line 61: Nit: capitalize "Lambda" in the last sentence.

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
+| `forwarderARN`         | When set, the plugin will automatically subscribe the Datadog Forwarder to the functions' log groups. Do not set `forwarderARN` when `extensionLayerVersion` is set.|
+| `flushMetricsToLogs` | Send custom metrics using Cloudwatch logs with the Datadog Forwarder Lambda function (recommended). Defaults to `true`. If you disable this parameter, it's required to set `apiKey` (or `apiKMSKey`). |
```
nhinsch, line 63: Nit: Cloudwatch -> CloudWatch

README.md
```diff
@@ -58,11 +58,12 @@ To further configure your Datadog construct, use the following custom parameters
...
+| `site`               | Set which Datadog site to send data. This is only used when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com` and `ddog-gov.com`. The default is `datadoghq.com`. |
+| `apiKey`             | Datadog API Key, only needed when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set. For more information about getting a Datadog API key, see the [API key documentation][8]. |
+| `apiKMSKey`          | Datadog API Key encrypted using KMS. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set, and you are using KMS encryption. |
 | `enableDDTracing`    | Enable Datadog tracing on your Lambda functions. Defaults to `true`. When enabled, it's required to set the `forwarderARN` parameter. |
```
nhinsch, line 67: This makes sense for now, but in the future, when the extension supports tracing, we should revisit this parameter. Can you put a story in the backlog on Jira so we don't forget?

lib/datadog.ts
```diff
@@ -63,3 +72,32 @@ export class Datadog extends cdk.Construct {
...
+function validateProps(props: DatadogProps) {
+  const siteList: string[] = ["datadoghq.com", "datadoghq.eu", "us3.datadoghq.com", "ddog-gov.com"];
+  if (props.apiKey !== undefined && props.apiKMSKey !== undefined) {
+    throw new Error("Both `apiKey` and `apiKMSKey` cannot be set.");
+  }
+
+  // If the extension is used, metrics will be submitted via the extension.
+  if (props.extensionLayerVersion !== undefined) {
+    props.flushMetricsToLogs = false;
+  }
```
nhinsch, line 85: Hmm, I would take this piece out of `validateProps()` and put it back to where we decide how to send metrics. The reason is that this is basically changing the props, which goes a step beyond validating them. If I just want to validate something, I wouldn't expect it to change.

README.md
nhinsch: Also -- are we checking this somewhere? Do we need to add this to the validate function?

lib/transport.ts
```diff
@@ -65,10 +61,10 @@ export class Transport {
       if (this.site !== undefined && this.flushMetricsToLogs === false) {
         lam.addEnvironment(siteURLEnvVar, this.site);
       }
-      if (this.apiKey !== undefined && this.flushMetricsToLogs === false) {
+      if (this.apiKey !== undefined) {
         lam.addEnvironment(apiKeyEnvVar, this.apiKey);
       }
-      if (this.apiKMSKey !== undefined && this.flushMetricsToLogs === false) {
+      if (this.apiKMSKey !== undefined) {
         lam.addEnvironment(apiKeyKMSEnvVar, this.apiKMSKey);
```
zARODz11z, line 68: These changes make sure that if an API key is provided then we will always add it as an env variable

lib/datadog.ts
```diff
@@ -63,3 +73,34 @@ export class Datadog extends cdk.Construct {
...
+      "Warning: Invalid site URL. Must be either datadoghq.com, datadoghq.eu, us3.datadoghq.com, or ddog-gov.com.",
+    );
+  }
+
+  if (props.apiKey === undefined && props.apiKMSKey === undefined && props.flushMetricsToLogs === false) {
+    throw new Error("When `flushMetricsToLogs` is false, `apiKey` or `apiKMSKey` must also be set.");
+  }
+
+  // TODO: Refactor this error when Extension layer officially supports tracing.
+  if (props.enableDDTracing === true && props.forwarderARN === undefined) {
```
nhinsch, line 94: After thinking about this more, I think we should not add this check. Otherwise it will be impossible to use the extension for tracing until we update this. So, let's leave this out. Sorry for the confusion.

nhinsch (approved):

PR #30, opened 2021-04-16
Ensure log group subscription filter resource has stable id

mweberxyz:
### What does this PR do?

This PR calls ensures the log filter subscription resource id does not change on subsequent cdk deploys.

### Motivation

Every time I deploy my app utilizing datadog-cdk-constructs, I see that the log subscription filters are deleted and re-added by CloudFormation. This due to references being passed in as the functionArn to generateForwaderConstructId, and the reference ID numbers are not stable across synth operations. See below for back-to-back cdk diff operations:

CleanShot 2021-04-15 at 20 59 10

### Testing Guidelines

An additional unit test was added that shows two identical -- but independent -- stacks produce the same name of the subscription filter, but a third stack with a differently named resource does not.

### Additional Notes

The implementation of the existing unit tests inlined the actual id generation algorithm, so I modified these tests to instead introspect the CDK data structures to confirm everything is working.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Misc (test refactoring)

### Check all that apply

### DIFF
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -187,7 +187,7 @@ describe("addForwarder", () => {
     const [differentStackSubscription] = findDatadogSubscriptionFilters(differentStack);
 
     expect(initialStackSubscription.id).toEqual(identicalStackSubscription.id);
-    expect(initialStackSubscription).not.toEqual(differentStackSubscription.id);
+    expect(initialStackSubscription.id).not.toEqual(differentStackSubscription.id);
   });
 });
 describe("applyLayers", () => {
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -35,11 +35,11 @@ describe("addLambdaFunctions", () => {
     datadogCdk.addLambdaFunctions([nodeLambda]);
     datadogCdk.addLambdaFunctions([pythonLambda]);
 
-    const [nodeLambdaSubscriptionFilter] = findDatadogSubscriptionFilters(nodeLambda);
-    const [pythonLambdaSubscriptionFilter] = findDatadogSubscriptionFilters(pythonLambda);
-    expect(nodeLambdaSubscriptionFilter).toBeDefined();
-    expect(pythonLambdaSubscriptionFilter).toBeDefined();
-    expect(nodeLambdaSubscriptionFilter.destinationArn).toEqual(pythonLambdaSubscriptionFilter.destinationArn);
+    const nodeLambdaSubscriptionFilters = findDatadogSubscriptionFilters(nodeLambda);
+    const pythonLambdaSubscriptionFilters = findDatadogSubscriptionFilters(pythonLambda);
+    expect(nodeLambdaSubscriptionFilters).toHaveLength(1);
+    expect(pythonLambdaSubscriptionFilters).toHaveLength(1);
+    expect(nodeLambdaSubscriptionFilters[0].destinationArn).toEqual(pythonLambdaSubscriptionFilters[0].destinationArn);
   });
 
   it("Throws an error when a customer redundantly calls the addLambdaFunctions function on the same lambda function(s) and forwarder", () => {
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -1,30 +1,11 @@
 import * as lambda from "@aws-cdk/aws-lambda";
-import { CfnSubscriptionFilter } from "@aws-cdk/aws-logs";
-
 import * as cdk from "@aws-cdk/core";
 import "@aws-cdk/assert/jest";
-import { addForwarder } from "../src/forwarder";
 import { Datadog, logForwardingEnvVar, enableDDTracingEnvVar, injectLogContextEnvVar } from "../src/index";
 import { JS_HANDLER_WITH_LAYERS, DD_HANDLER_ENV_VAR, PYTHON_HANDLER } from "../src/redirect";
-const SUBSCRIPTION_FILTER_PREFIX = "DatadogSubscriptionFilter";
+import { findDatadogSubscriptionFilters } from "./test-utils";
 
-function findDatadogSubscriptionFilters(stack: cdk.Stack) {
-  return stack.node
-    .findAll()
-    .filter((construct) => construct.node.id.startsWith(SUBSCRIPTION_FILTER_PREFIX))
-    .map((construct) => {
-      const cfnSubscriptionFilter = construct.node.children.find(
-        (child) => child instanceof CfnSubscriptionFilter,
-      ) as CfnSubscriptionFilter;
-
-      return {
-        id: construct.node.id,
-        destinationArn: cfnSubscriptionFilter.destinationArn,
-      };
-    });
-}
-
-describe("addForwarder", () => {
+describe("addLambdaFunctions", () => {
   it("Subscribes the same forwarder to two different lambda functions via separate addLambdaFunctions function calls", () => {
     const app = new cdk.App();
     const stack = new cdk.Stack(app, "stack", {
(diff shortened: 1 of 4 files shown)


zARODz11z:
@mweberxyz Thanks for your contribution I will take a look

mweberxyz:
Thanks @zARODz11z, to contrast with my screenshot in the description of the problem, here is the result of two consecutive `cdk diff` in my primary project when incorporating my changes. Note each run attempts to set the resource ID to the same value (and then future diff/deploy from that point forward would ensure the subscription filters are not re-created):

CleanShot 2021-04-16 at 15 37 12

zARODz11z:
For context the reason why we have that subscription filter id logic is because we had issues with duplicate constructs in the same construct tree. So I may have to do some research/testing to see if these changes don't re-introduce that problem.

zARODz11z:
I took a look and even manually tested your changes, they look solid. For the sake of getting a fresh look at it, I am going to wait until Monday to provide any feedback/approve these changes.

test/index.spec.ts
```diff
@@ -48,16 +53,11 @@ describe("addForwarder", () => {
...
-    );
-    const pythonLambdaLogGroupSubscriptionFilterName = createSubscriptionFilterName(
-      pythonLambda.functionArn,
-      "forwarder-arn",
-    );
-    expect(nodeLambda.logGroup.node.tryFindChild(nodeLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
-    expect(pythonLambda.logGroup.node.tryFindChild(pythonLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
+
+    const subscriptionFilters = findDatadogSubscriptionFilters(stack);
+    expect(nodeLambda.logGroup.node.tryFindChild(subscriptionFilters[0].id)).not.toBeUndefined();
```
zARODz11z, line 58: I like how you refactored the tests but how can you be sure that index 0 of the `subscriptionFilters` array will always be the subscription filter associated with `nodeLambda` and subsequently how index 1 will be the filter associated with `pythonLambda`

test/index.spec.ts
```diff
@@ -78,16 +78,11 @@ describe("addForwarder", () => {
       handler: "hello.handler",
     });
     addForwarder(stack, [nodeLambda, pythonLambda], "forwarder-arn");
```
zARODz11z, line 80: I noticed that the `Subscribes the same forwarder to two different lambda functions via separate addLambdaFunctions function calls` test uses `datadogCdk.addLambdaFunctions([nodeLambda]);` and `datadogCdk.addLambdaFunctions([pythonLambda]);` to add the subscriptions however this test directly uses `addForwarder`. Could you update the tests to use the `addForwarder` function instead of the `datadogCdk.addLambdaFunctions()` way? 

This way the convention in the file is the same and so the `addForwarder` set of tests really only test the `addForwarder` function and not all the other codepaths that come with using `datadogCdk.addLambdaFunctions()`

test/index.spec.ts
```diff
@@ -48,16 +53,11 @@ describe("addForwarder", () => {
     });
     datadogCdk.addLambdaFunctions([nodeLambda]);
```
zARODz11z, line 54: see related [comment](https://github.com/DataDog/datadog-cdk-constructs/pull/30/files#r616051414)

test/index.spec.ts
```diff
@@ -48,16 +53,11 @@ describe("addForwarder", () => {
...
     datadogCdk.addLambdaFunctions([pythonLambda]);
```
zARODz11z, line 55: see related [comment](https://github.com/DataDog/datadog-cdk-constructs/pull/30/files#r616051414)

test/index.spec.ts
```diff
@@ -158,16 +153,41 @@ describe("addForwarder", () => {
     });
     datadogCdk.addLambdaFunctions([nodeLambda]);
```
zARODz11z, line 154: see related [comment](https://github.com/DataDog/datadog-cdk-constructs/pull/30/files#r616051414)

test/index.spec.ts
```diff
@@ -158,16 +153,41 @@ describe("addForwarder", () => {
...
     datadogCdk2.addLambdaFunctions([pythonLambda]);
```
zARODz11z, line 155: see related [comment](https://github.com/DataDog/datadog-cdk-constructs/pull/30/files#r616051414)

zARODz11z (changes requested):
@mweberxyz Overall this is a great addition however I left some requests for changes and questions before merging.

test/index.spec.ts
```diff
@@ -1,24 +1,29 @@
...
+
+      return {
+        id: construct.node.id,
+        destinationArn: cfnSubscriptionFilter.destinationArn,
+      };
+    });
 }
+
 describe("addForwarder", () => {
   it("Subscribes the same forwarder to two different lambda functions via separate addLambdaFunctions function calls", () => {
```
zARODz11z, line 28: addLambdaFunctions -> addForwarder

test/index.spec.ts
```diff
@@ -78,16 +78,11 @@ describe("addForwarder", () => {
...
-    expect(nodeLambda.logGroup.node.tryFindChild(nodeLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
-    expect(pythonLambda.logGroup.node.tryFindChild(pythonLambdaLogGroupSubscriptionFilterName)).not.toBeUndefined();
+
+    const subscriptionFilters = findDatadogSubscriptionFilters(stack);
+    expect(nodeLambda.logGroup.node.tryFindChild(subscriptionFilters[0].id)).not.toBeUndefined();
+    expect(pythonLambda.logGroup.node.tryFindChild(subscriptionFilters[1].id)).not.toBeUndefined();
+    expect(subscriptionFilters[0].destinationArn).toEqual(subscriptionFilters[1].destinationArn);
   });
 
   it("Throws an error when a customer redundantly calls the addLambdaFunctions function on the same lambda function(s) and forwarder", () => {
```
zARODz11z, line 88: disregard this [comment](https://github.com/DataDog/datadog-cdk-constructs/pull/30/files#r616051414) for this test

test/index.spec.ts
mweberxyz: Good question. The answer was "it works today on my machine" but for sanity's sake I factored it out to look at the children of the lambda's themselves.

test/index.spec.ts
mweberxyz: I didn't want to get too deep into reworking the tests, but seems best to move all addForwarder specific tests to the forwarder.spec.ts, and factor out the "search the cdk nodes" logic to be reused by both index.spec and forwarder.spec

Let me know how you feel about the most recent commit!

mweberxyz:
Hey there @zARODz11z let me know how you feel about the most recent commit. I think I understood what you were getting at.

I added an export to forwarder.ts to avoid redefining the constant in the tests, then  moved the tests that invoke addForwarder specifically to forwarder.spec and factored out the CDK tree introspection to use it both in index.spec and forwarder.spec.

test/test-utils.ts
```diff
@@ -0,0 +1,21 @@
...
+export const findDatadogSubscriptionFilters = (baseConstruct: cdk.Construct) => {
+  return baseConstruct.node
+    .findAll()
+    .filter((construct) => construct.node.id.startsWith(SUBSCRIPTION_FILTER_PREFIX))
+    .map((construct) => {
+      const cfnSubscriptionFilters = construct.node.children.filter(
+        (child) => child instanceof CfnSubscriptionFilter,
+      ) as CfnSubscriptionFilter[];
+
+      expect(cfnSubscriptionFilters).toHaveLength(1);
```
zARODz11z, line 14: Is this because you are only ever expecting to get 1 subscription filter per call to `findDatadogSubscriptionFilters`

zARODz11z (commented):
@mweberxyz  This looks good to me, I left 1 question.

test/test-utils.ts
mweberxyz: Correct, I am asserting that the baseConstruct only has a single subscription filter.

Because a future test might want to ensure that a given stack as the baseConstruct has 2 subscription filters (for example, with 2 forwarder arns) I think it makes sense to gate this with a typecheck -- any only assert that there is a single subscription filter if the baseConstruct is a Function. I will push this shortly.

mweberxyz: I pushed the change but actually.. hold off on reviewing. I will make this a bit clearer.

mweberxyz:
Alrighty @zARODz11z I think it's as clear as possible at this point, let me know what you think.

zARODz11z (approved):
This looks good to me, thanks for the contribution!

PR #37, opened 2021-04-28
Improve Release Script

zARODz11z:
### What does this PR do?
This PR attempts to make the release script more rigid. For instance:
* I strengthened the regex to only allow digits. (When someone would call the script using `v3.1.4` it would fail because many of the commands later in the script did not expect there to be a `v` just strictly the `major.minor.patch`)
* Before going any further in the script we check to make sure the user has all the necessary terminal commands as that would be a silly way to end the script. One instance I am trying to avoid is the JS package being published but then the Python one fails because they dont have `pip` or `python3`. I would like to minimize the chance that the script fails in a critical part like the publishing to NPM and PyPi.
* Before publishing the artifacts we make sure that they actually exist. This again stops the situation of one package being published but the other failing and being out of sync.

### Motivation
The script had a lot of room to fail. Failing on a release script could cause some problems therefore before doing any publishing we would like to do our best to ensure everything will run smoothly before getting to the critical parts of the script. 

### Testing Guidelines
I manually ran the script end to end on my test pypi and npm accounts (I deleted the packages right after). I asserted that the pitfalls where the script could easily fail before have checks. The release process should be smooth from now on.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/publish_prod.sh
+++ b/scripts/publish_prod.sh
@@ -30,6 +30,22 @@ else
     VERSION=$1
 fi
 
+if ! [ -x "$(command -v yarn)" ]; then
+  echo 'Error: yarn is not installed.'
+  exit 1
+fi
+if ! [ -x "$(command -v pip)" ]; then
+  echo 'Error: pip is not installed.'
+  exit 1
+fi
+if ! [ -x "$(command -v python3)" ]; then
+  echo 'Error: python3 is not installed.'
+  exit 1
+fi
+
+# Make sure dependencies are installed before proceeding
+yarn
+
 read -p "Do you have the PyPI and npm login credentials for the Datadog account (y/n)?" CONT
 if [ "$CONT" != "y" ]; then
     echo "Exiting"
--- a/scripts/publish_prod.sh
+++ b/scripts/publish_prod.sh
@@ -64,7 +64,7 @@ fi
 
 echo "Bumping the version number and committing the changes"
 if git log --oneline -1 | grep -q "chore(release):"; then
-    echo "Create a new commit before attempting to release. Be sure to not include 'chore(release):' in the commit message, aborting"
+    echo "Create a new commit before attempting to release. Be sure to not include 'chore(release):' in the commit message, this means if the script prematurely ended previously without publishing you may need to 'git reset --HARD' to a previous commit before trying again, aborting"
     exit 1
 else
     yarn standard-version --release-as $VERSION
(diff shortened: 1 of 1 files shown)


scripts/publish_prod.sh
```diff
@@ -23,7 +23,7 @@ CURRENT_VERSION=$(node -pe "require('./package.json').version")
 if [ -z "$1" ]; then
     echo "Must specify a desired version number"
     exit 1
-elif [[ ! $1 =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
+elif [[ ! $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
zARODz11z, line 26: NOTE `3.2.222222` is allowed, should we stop this? maybe only allow each num 0-9?

scripts/publish_prod.sh
```diff
@@ -41,24 +57,50 @@ rm -rf dist;
...
-yarn login
+if git rev-parse "v${VERSION}" >/dev/null 2>&1; then
+    echo "tag v${VERSION} already exists, aborting"
+    exit 1
+fi
 
 echo "Bumping the version number and committing the changes"
 if git log --oneline -1 | grep -q "chore(release):"; then
-    echo "Create a new commit before attempting to release. Be sure to not include 'chore(release):' in the commit message, aborting"
+    echo "Create a new commit before attempting to release. Be sure to not include 'chore(release):' in the commit message. This means if the script previously prematurely ended without publishing you may need to 'git reset --hard' to a previous commit before trying again, aborting"
```
nhinsch, line 72: I'm not completely sure I understand the meaning/purpose of this message, can you explain it a bit?

zARODz11z: @nhinsch yeah so on line 75 we use `yarn standard-version --release-as $VERSION` to bump the version number and commit the changes. As an example if someone ran the release script wanting to release version 1.0.0 then a new commit would be added with the message: `chore(release): 1.0.0`. 

So despite all the safeguards that I put in place, if for some reason the script ended prematurely after line 75 then the user wouldn't be able to run the script from beginning to end because of this logic on line 71. We want to ensure 2 releases are not made right after another, there should be some other commits between them. If someone found themselves in this situation I figured the message may help them understand.

nhinsch (approved):

PR #39, opened 2021-05-11
[Datadog CDK] Add function for subscribing additional log groups

jcstorms1:
### What does this PR do?

- Adds a new function `addForwarderToLogGroups` so users can subscribe to any non-lambda function based log groups.
- Adds new unit tests
- Now allows the forwarder and extension to be enabled

### Motivation

https://datadoghq.atlassian.net/browse/SLS-1123

### Testing Guidelines

Unit tests and deployed to sandbox

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


package.json
```diff
@@ -46,20 +49,23 @@
     "jsii-pacmak": "^1.26.0",
     "json-schema": "^0.3.0",
     "prettier": "^2.2.1",
-    "projen": "^0.16.42",
+    "projen": "^0.17.84",
```
zARODz11z, line 52: It seems like this may be the culprit for why the additional projen stuff was added, I would say we should stay at "^0.16.42" for now

package.json
```diff
@@ -70,7 +76,7 @@
   ],
   "main": "lib/index.js",
   "license": "Apache-2.0",
-  "version": "0.2.3",
+  "version": "0.0.0",
```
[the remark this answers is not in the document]
jcstorms1: Fixed with revert

zARODz11z:
Keep in mind that this repo is published to npm and pypi. There doesn't seem to be any errors from JSII but whenever I make changes to the CDK I usually like to test the package that is going to be published to npm and pypi before merging just as another confidence boost. It may be more work at the moment but you wont regret it if you find out something was broken before merging. You can use this project https://github.com/DataDog/cloud-integrations-lambdas/tree/master/lambda-cdk-constructs-demo . Instead of pulling the live Datadog CDK package you can use local versions and test your changes manually. I think of it in the mind of a outside user, like if you were an customer would this all make sense. Then in that sense you can interact with the new API you created in the way a user would.

README.md
```diff
@@ -151,7 +152,9 @@ Add tags to your constructs. We recommend setting an `env` and `service` tag to
...
-The Datadog CDK construct takes in a list of lambda functions and installs the Datadog Lambda Library by attaching the Lambda Layers for [Node.js][2] and [Python][1] to your functions. It redirects to a replacement handler that initializes the Lambda Library without any required code changes. Additional configurations added to the Datadog CDK construct will also translate into their respective environment variables under each lambda function (if applicable / required).
+The Datadog CDK construct takes in a list of lambda functions and installs the Datadog Lambda Library by attaching the Lambda Layers for [Node.js][2] and [Python][1] to your functions. It redirects to a replacement handler that initializes the Lambda Library without any required code changes. Additional configurations added to the Datadog CDK construct will also translate into their respective environment variables under each lambda function (if applicable / required). 
+
+While Lambda function based log groups are handled by the `addForwarder` function automatically, the construct has an additional function `addForwarderToLogGroups` which will subscribe the forwarder to any additional log groups of your choosing.
```
zARODz11z, line 157: The only API's that a user can interact with are:
- addLambdaFunctions
- addForwarderToLogGroups
so I am wondering if it would make more sense to say: "While Lambda function based log groups are handled by the `addLambdaFunctions` function automatically,"  

since `addForwarder` is an internal function in which the user does not interact with.

src/datadog.ts
```diff
@@ -78,9 +86,14 @@ export class Datadog extends cdk.Construct {
         this.props.extensionLayerVersion,
       );
       redirectHandlers(lambdaFunctions, addLayers);
+
       if (this.props.forwarderArn !== undefined) {
-        log.debug(`Adding log subscriptions using provided Forwarder ARN: ${this.props.forwarderArn}`);
-        addForwarder(this.scope, lambdaFunctions, this.props.forwarderArn);
+        if (this.props.extensionLayerVersion !== undefined) {
+          log.debug(`Skipping adding subscriptions to the lambda functions since the extension is enabled`);
```
zARODz11z, line 92: Nit: I think the subscriptions get added to the lambda log groups and not directly to the function. 
Maybe this is clearer:
`Skipping adding subscriptions to the lambda log groups since the extension is enabled`

src/datadog.ts
```diff
@@ -90,6 +103,14 @@ export class Datadog extends cdk.Construct {
       this.transport.applyEnvVars(lambdaFunctions);
     }
   }
+
+  public addForwarderToLogGroups(logGroups: logs.LogGroup[]) {
+    if (this.props.forwarderArn !== undefined) {
+      addForwarderToLogGroup(this.scope, logGroups, this.props.forwarderArn);
```
zARODz11z, line 109: Nit: I feel like its strange that we have a `addForwarderToLogGroups` (plural) and a `addForwarderToLogGroup` (singular) but they both can take in multiple log groups. I feel like the call to `addForwarderToLogGroup` (line 109) is a bit misleading maybe we can think of a different name. 

I know the user wont ever know that we call another function inside the `addForwarderToLogGroups` function but it may help with code readability when developers look back at this.

src/datadog.ts
```diff
@@ -90,6 +103,14 @@ export class Datadog extends cdk.Construct {
...
+    } else {
+      log.debug("Forwarder ARN not provided, no log group subscriptions will be added");
```
zARODz11z, line 111: Nit: This debug statement is exactly the same as the one in the `addLambdaFunctions` function [here](https://github.com/DataDog/datadog-cdk-constructs/blob/d5b9c0adecc1d3b165da3880fa517c5d6b4e63ca/src/datadog.ts#L98), it might be useful to keep the core message the same but differ them slightly so that you can distinguish what code path they came from if you were to look at the debug logs.

src/datadog.ts
```diff
@@ -90,6 +103,14 @@ export class Datadog extends cdk.Construct {
       this.transport.applyEnvVars(lambdaFunctions);
     }
   }
+
+  public addForwarderToLogGroups(logGroups: logs.LogGroup[]) {
```
zARODz11z, line 107: Maybe we should consider changing this to `addForwarderToNonLambdaLogGroups ` or something shorter but include that its made for non lambda log groups. We have the addLambdaFunctions function that takes care of lambda based log groups automatically but if someone didn't read the docs right they could see the `addForwarderToLogGroups` function and think its meant for their lambda function log groups and go through unnecessary steps with the DD CDK

zARODz11z:
Since you have your manual testing environment setup it might be worth while to try to set up a apigateway endpoint and pass its log group into the current `addForwarderToLogGroups` function then actually hit the apitgatway endpoint and see if forwarder sends up any of those apigatway logs to the Datadog app (after all I think that was the entire point of this work right?)

src/forwarder.ts
```diff
@@ -34,14 +34,17 @@ function generateSubscriptionFilterName(functionUniqueId: string, forwarderArn:
   return subscriptionFilterName;
 }
 
-export function addForwarder(scope: cdk.Construct, lambdaFunctions: lambda.Function[], forwarderArn: string) {
+function getForwarder(scope: cdk.Construct, forwarderArn: string) {
```
zARODz11z, line 37: I like how you moved the forwarder stuff into a separate function, it keeps the code cleaner and easier to read

src/sample/index.ts
```diff
@@ -21,33 +23,60 @@ export class ExampleStack extends cdk.Stack {
...
+
     const hello1 = new lambda.Function(this, "HelloHandler1", {
       runtime: lambda.Runtime.NODEJS_10_X,
       code: lambda.Code.fromAsset("./src/sample/lambda"),
       handler: "hello.handler",
     });
 
+    const restLogGroup1 = new LogGroup(this, "restLogGroup1");
+    new LambdaRestApi(this, "rest-test1", {
+      handler: hello,
```
zARODz11z, line 43: is this supposed to be `hello1`?

src/sample/index.ts
```diff
@@ -21,33 +23,60 @@ export class ExampleStack extends cdk.Stack {
...
     const hello2 = new lambda.Function(this, "HelloHandler2", {
       runtime: lambda.Runtime.PYTHON_3_7,
       code: lambda.Code.fromAsset("./src/sample/lambda"),
       handler: "hello_py.handler",
       tracing: lambda.Tracing.ACTIVE,
     });
 
+    const restLogGroup2 = new LogGroup(this, "restLogGroup2");
+    new LambdaRestApi(this, "rest-test2", {
+      handler: hello,
```
zARODz11z, line 58: same thing is this supposed to be `hello2`?

src/sample/index.ts
```diff
@@ -21,33 +23,60 @@ export class ExampleStack extends cdk.Stack {
...
+      deployOptions: {
+        accessLogDestination: new LogGroupLogDestination(restLogGroup2),
+      },
+    });
+
     const datadogCDK = new Datadog(this, "Datadog", {
       nodeLayerVersion: 39,
       pythonLayerVersion: 24,
-      forwarderArn: "<forwarder_ARN>",
+      forwarderArn: "arn:aws:lambda:sa-east-1:601427279990:function:datadog-forwarder-staging-org-2",
```
zARODz11z, line 67: We can probably just remove the actual forwarder arn and leave it with the place holder

zARODz11z: I know you never added an api key but FYI if you ever do I would set up an env variable for in your terminal so that in the future you dont accidentally commit an api key.

src/sample/index.ts
```diff
@@ -21,33 +23,60 @@ export class ExampleStack extends cdk.Stack {
...
     });
     datadogCDK.addLambdaFunctions([hello, hello1, hello2]);
+    datadogCDK.addForwarderToNonLambdaLogGroups([restLogGroup, restLogGroup1, restLogGroup2]);
   }
 }
 
 const app = new cdk.App();
-const stack = new ExampleStack(app, "ExampleStack");
+const env = { account: "601427279990", region: "sa-east-1" };
+const stack = new ExampleStack(app, "storms-test-stack", { env: env });
```
zARODz11z, line 80: We can probably change `storms-test-stack` -> `ExampleDatadogCdkStack`

test/datadog.spec.ts
```diff
@@ -193,4 +159,57 @@ describe("addCdkTag", () => {
...
+      site: "datadoghq.com",
+      forwarderArn: "forwarder-arn",
+      apiKey: "1234",
+    });
+
+    datadogCdk.addLambdaFunctions([hello1]);
+    expect(stack).not.toHaveResource("AWS::Logs::SubscriptionFilter");
+  });
+
+  it("Does call the addForwarderToLogs function when the extension is enabled", () => {
```
zARODz11z, line 190: I think `addForwarderToLogs` might be out of date in this message

test/forwarder.spec.ts
```diff
@@ -151,4 +154,97 @@ describe("Forwarder", () => {
     expect(stackOneSubscriptions[0].destinationArn).not.toEqual(stackTwoSubscriptions[0].destinationArn);
     expect(stackOneSubscriptions[0].id).not.toEqual(stackTwoSubscriptions[0].id);
   });
+
+  it("Subscribes the forwarder to a log group via the addForwarderToLogGroups functioñ", () => {
```
zARODz11z, line 158: Nit: functioñ -> function

zARODz11z (approved):
Looks good! Left some minor comments above

PR #40, opened 2021-05-11
Anti-tamper to project files governed by Projen

zARODz11z:
### What does this PR do?
Adds a diff step in the build workflow that will check if the developer made changes to the project files governed by projen directly instead of going through the projenrc file.

### Motivation
The project is set up with Projen to control most of the configuration files, ie. the package.json, gitignore, ... are generated by projen. Therefore if someone directly edits the package.json file and committed that change then the next time someone ran `npx projen` the package.json would change since the projenrc file has a direct link to how the project configuration files are synthesized.

### Testing Guidelines
I manually tested this feature by adding intentional commits that alterned the package.json file directly instead of going through the projenrc file. 

[Here](https://github.com/DataDog/datadog-cdk-constructs/runs/2561116625) you can see that the anti-tamper check failed because the package.json "name": "datadog-cdk-constructs" was directly changed to "name": "datadoge-cdk-constructs".

### Additional Notes
The way this works is that in the anti-tamper step we run `npx projen` which synthesizes the project files according to the `projenrc.js` file and if there is a diff using `git diff --exit-code` then we know that the developer directly edited the configuration file without going through the `projenrc.js` file. 

The reason why we know is if the user had implemented their changes into the projenrc.js file then there would have been no differences when running `npx projen`.  This idea was taken from Projen's included github workflows. You can check one out [here](https://github.com/projen/projen/blob/8efe432cac121b642d3ba20dc344f5f9dcb23bbf/.github/workflows/build.yml#L36).

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


DarcyRaynerDD (approved):
LGTM

PR #50, opened 2021-08-16
Adding support for DD_API_KEY_SECRET_ARN

CNorskov:
### What does this PR do?
This PR adds support for DD_API_KEY_SECRET_ARN for the lambda extension.

### Motivation
We have the datadog api key stored in secrets manager, and want to use the datadog cdk construct for easier deployment and management of our lambda functions.

### Testing Guidelines
Only unit tests.

### Additional Notes
None that I can think of

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR impacts documentation, and it has been updated (or a ticket has been logged)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ const datadog = new Datadog(this, "Datadog", {
   site: "<SITE>",
   apiKey: "{Datadog_API_Key}",
   apiKmsKey: "{Encrypted_Datadog_API_Key}",
+  apiKeySecretArn: "{Secret_ARN_For_Datadog_API_Key}",
   enableDatadogTracing: <BOOLEAN>,
   enableDatadogLogs: <BOOLEAN>,
   injectLogContext: <BOOLEAN>,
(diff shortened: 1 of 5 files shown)


nhinsch:
Hi @CNorskov! Thank you for your contribution! 😄 

As you noticed, the API key is required in two situations:
- When using the extension
- When sending metrics to Datadog synchronously directly from the Datadog Lambda Library (without the use of the extension). This is no longer recommended but still supported. To enable this, `flushMetricsToLogs` is set to `false`.

Right now, the extension supports the `DD_API_KEY_SECRET_ARN`, as does our Python library. However, our Node library does not support this environment variable.

Therefore, in order to merge this pull request, we will need to make sure that if `flushMetricsToLogs` is set to false, the user must provide a `apiKmsKey` or `apiKey`. `apiKeySecretArn` is not yet supported for this use case because of the missing feature in the Node library.

However, we can allow `apiKeySecretArn` for the extension use case.

The alternative would be to add support for `DD_API_KEY_SECRET_ARN` to the Node library, but this would require more work and testing than the solution I outlined above.

Does that all make sense?

CNorskov:
> Hi @CNorskov! Thank you for your contribution! 😄
> 
> As you noticed, the API key is required in two situations:
> 
> * When using the extension
> * When sending metrics to Datadog synchronously directly from the Datadog Lambda Library (without the use of the extension). This is no longer recommended but still supported. To enable this, `flushMetricsToLogs` is set to `false`.
> 
> Right now, the extension supports the `DD_API_KEY_SECRET_ARN` environment variable, as does our Python library. However, our Node library does not support this environment variable.
> 
> Therefore, in order to merge this pull request, we will need to make sure that if `flushMetricsToLogs` is set to false, the user must provide `apiKmsKey` or `apiKey`. `apiKeySecretArn` is not yet supported for this use case because of the missing feature in the Node library.
> 
> However, we can allow `apiKeySecretArn` for the extension use case.
> 
> The alternative would be to add support for `DD_API_KEY_SECRET_ARN` to the Node library, but this would require more work and testing than the solution I outlined above.
> 
> Does that all make sense?

Hi @nhinsch 

Thanks for the detailed comment.
I guess it makes sense. Unfortunately the suggested changes renders it useless for our usecase. 
And my typescript skills does not extend to update the Node library.

So I'll probably close this PR.

Best regards,
Christian.

PR #54, opened 2021-09-10
Support for NodejsFunction construct

ajsaavedra:
### What does this PR do?
Enables the datadog cdk construct to accept lambda functions instantiated with the [aws-cdk-lambda-nodejs](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda-nodejs.NodejsFunction.html) construct.

### Motivation

Part of ongoing work to update the construct and enabling Nodejs and Python constructs 
https://datadoghq.atlassian.net/browse/SLS-1420

### Testing Guidelines

Updated the `index.spec.ts` to test for when users use a mix of Lambda.Function or NodejsFunction types.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/datadog.ts
+++ b/src/datadog.ts
@@ -61,19 +61,7 @@ export class Datadog extends cdk.Construct {
     );
   }
 
-  public addLambdaFunctions(lambdaFunctions: lambda.Function[]) {
-    if (this.props !== undefined && lambdaFunctions.length > 0) {
-      this.handleLambdaConstruct(lambdaFunctions);
-    }
-  }
-
-  public addLambdaNodejsFunctions(lambdaNodejsFunctions: lambdaNodejs.NodejsFunction[]) {
-    if (this.props !== undefined && lambdaNodejsFunctions.length > 0) {
-      this.handleLambdaConstruct(lambdaNodejsFunctions);
-    }
-  }
-
-  private handleDatadogConstructProps() {
+  public addLambdaFunctions(lambdaFunctions: (lambda.Function | lambdaNodejs.NodejsFunction)[]) {
     let addLayers = this.props.addLayers;
     let enableDatadogTracing = this.props.enableDatadogTracing;
     let injectLogContext = this.props.injectLogContext;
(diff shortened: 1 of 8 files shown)


src/sample/index.ts
```diff
@@ -18,15 +19,21 @@ export class ExampleStack extends cdk.Stack {
 
     // user's lambda function
     const hello = new lambda.Function(this, "HelloHandler", {
-      runtime: lambda.Runtime.NODEJS_10_X,
+      runtime: lambda.Runtime.NODEJS_14_X,
```
nhinsch, line 22: Good that we're switching this from Node 10 to Node 14, since Node 10 is no longer supported. However, I still see some other uses of Node 10 in this file, can we switch those as well?

src/sample/index.ts
```diff
@@ -18,15 +19,21 @@ export class ExampleStack extends cdk.Stack {
 
     // user's lambda function
```
nhinsch, line 20: Nit: I know this was already there, but I'm not sure this comment ("user's lambda function") makes sense given that there is more than one Lambda function defined in this sample app. I would 🔪 it.

src/sample/index.ts
```diff
@@ -62,15 +69,15 @@ export class ExampleStack extends cdk.Stack {
...
+      nodeLayerVersion: 62,
+      extensionLayerVersion: 10,
       forwarderArn: "<forwarder_ARN>",
       enableDatadogTracing: true,
       flushMetricsToLogs: true,
       apiKey: "1234",
       site: "datadoghq.com",
     });
-    datadogCDK.addLambdaFunctions([hello, hello1, hello2]);
+    datadogCDK.addLambdaFunctions([hello, helloNodeJS, hello1, hello2]);
```
nhinsch, line 80: I think these function names are potentially confusing. They were already this way before, but now that we're adding another one it's a good opportunity to clarify this. Just looking at the names, it's not obvious what makes each function different. I think we should re-evaluate what all of these functions are called so that it's more clear what each of them is for.

Another thing worth thinking about is that including "NodeJS" by itself in the function name might not communicate that this function's significance is that it is created with the NodeJSFunction construct. Someone might just think it's called "helloNodeJS" because it's using the Node runtime (as opposed to Python or something else)

test/index.spec.ts
```diff
@@ -144,6 +145,43 @@ describe("addLambdaFunctions", () => {
...
+        flushMetricsToLogs: true,
+        site: "datadoghq.com",
+        forwarderArn: "forwarder-arn",
+        apiKey: "1234",
+      });
+      datadogCdk.addLambdaFunctions([hello1, hello2]);
+    } catch (e) {
+      threwError = true;
+    }
+    expect(threwError).toBe(false);
```
nhinsch, line 183: I think this is a good test, but I think there is a slight gap between the title of the test versus what it is actually testing.

The title says "adds aws-lambda-nodejs NodejsFunctions along with aws-lambda Functions". In this case "add" means we are going to be applying the Datadog instrumentation to the function. However, the test actually seems to just check that calling this function with a `NodejsFunction` doesn't throw an error. That's good, but it doesn't actually prove that we added the Datadog instrumentation correctly. I would expand this test a bit to check that we actually applied the Datadog instrumentation. There are some other tests in this file that could give you some inspiration for this!

test/index.spec.ts
```diff
@@ -144,6 +145,43 @@ describe("addLambdaFunctions", () => {
...
+        nodeLayerVersion: 20,
+        pythonLayerVersion: 28,
+        addLayers: true,
+        enableDatadogTracing: false,
+        flushMetricsToLogs: true,
+        site: "datadoghq.com",
+        forwarderArn: "forwarder-arn",
+        apiKey: "1234",
+      });
+      datadogCdk.addLambdaFunctions([hello1, hello2]);
```
nhinsch, line 179: Similar to my comment on the sample app, are `hello1` and `hello2` the most descriptive names here?

nhinsch:
Nit: I would rename the PR to be a bit more specific with regard to what exactly we are enabling here. Maybe something like "Support NodejsFunction construct". The title of the PR is what goes into the commit history/changelog so it's worth making it as clear as possible.

test/index.spec.ts
ajsaavedra: Thanks for the tip! I updated the test to check that the stack with a NodejsFunction gets the updated layers.

src/sample/index.ts
ajsaavedra: Great idea! Although both functions would use a lot of the same code, I added some helper functions to make this distinction/available method clear.

src/datadog.ts
```diff
@@ -61,6 +62,18 @@ export class Datadog extends cdk.Construct {
...
+    if (this.props !== undefined && lambdaFunctions.length > 0) {
+      this.handleLambdaConstruct(lambdaFunctions);
+    }
+  }
+
+  public addLambdaNodejsFunctions(lambdaNodejsFunctions: lambdaNodejs.NodejsFunction[]) {
+    if (this.props !== undefined && lambdaNodejsFunctions.length > 0) {
+      this.handleLambdaConstruct(lambdaNodejsFunctions);
+    }
+  }
```
nhinsch, line 74: I would merge these back into one function for simplicity.

nhinsch (approved):

PR #57, opened 2021-09-14
Datadog CDK Construct Integration tests

ajsaavedra:
### What does this PR do?

This PR adds a base integration test for the aws-lambda construct

### Motivation

https://datadoghq.atlassian.net/browse/SLS-1186

### Additional Notes

Running the script `run_integration_tests.sh` will run cdk and synthesize the example stack under `integration_tests/lambda/` to create an AWS CloudFormation template.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


integration_tests/tsconfig.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+  "include": [
+    "stacks/**/*.ts",
+    "../version.json"
+  ],
+  "exclude": [
+    "../node_modules"
+  ]
+}
+
```
nhinsch, line 42: Do we need a separate tsconfig.json just for the integration tests?

integration_tests/test_lambda_stack_snapshot.json
```diff
@@ -0,0 +1,90 @@
...
+        "Handler": "/opt/nodejs/node_modules/datadog-lambda-js/handler.handler",
+        "Layers": [
+          "arn:aws:lambda:sa-east-1:464622532012:layer:Datadog-Node14-x:XXX",
+          "arn:aws:lambda:sa-east-1:464622532012:layer:Datadog-Extension:XXX"
+        ],
+        "Runtime": "nodejs14.x",
+        "Tags": [
+          {
+            "Key": "dd_cdk_construct",
+            "Value": "v0.4.0"
```
nhinsch, line 69: I bet this version number will change over time, causing the snapshot tests to fail. Can we filter it out?

integration_tests/test_lambda_stack_snapshot.json
```diff
@@ -0,0 +1,90 @@
...
+              ]
+            ]
+          }
+        ]
+      },
+      "Metadata": {
+        "aws:cdk:path": "ExampleDatadogStack/HelloHandler/ServiceRole/Resource"
+      }
+    },
+    "HelloHandler2E4FBA4D": {
```
nhinsch, line 37: Are these sequences of characters such as `2E4FBA4D` going to be the same every time?

nhinsch:
Can we get these integration tests running in GitHub, so they run every time we push a branch?

integration_tests/test_lambda_stack_snapshot.json
nhinsch: I see some other stuff that looks like it may (or may not) need to be filtered out as well -- worth double checking if you haven't already

integration_tests/tsconfig.json
nhinsch: (Maybe we do, just curious)

ajsaavedra: I believe so -- our integration test compiles some test typescript files and rather than passing in flags to the `tsc` command, this was the easier option.

integration_tests/test_lambda_stack_snapshot.json
ajsaavedra: I've added some more steps to normalize the cdk output, just to be safe

nhinsch:
I noticed in the test output (in GitHub but I assume locally as well) that we log the entire CloudFormation stack when running the test, even if it's successful. This isn't going to break anything but it could be annoying, especially if we start to add lots of different snapshot tests. Is it possible to make it so that we don't log the full stack?

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,89 @@
...
+
+set -e
+
+# To add new tests create a new ts file in the 'integration_tests/lambda' directory, append it as a js file to the STACK_CONFIGS array
+# as well as creating a name for the snapshots that will be compared in your test. Add those snapshot names to the TEST_SNAPSHOTS and CORRECT_SNAPSHOTS arrays.
+# Note: Each yml config, test, and correct snapshot file should be at the same index in their own array. e.g. All the files for the forwarder test are at index 0.
+#       In order for this script to work correctly these arrays should have the same amount of elements.
+STACK_CONFIGS=("lambda-stack.js" "lambda-nodejs-function-stack.js")
+TEST_SNAPSHOTS=("test_lambda_stack_snapshot.json" "test_lambda_nodejs_function_stack_snapshot.json")
+CORRECT_SNAPSHOTS=("correct_lambda_stack_snapshot.json" "correct_lambda_nodejs_function_stack_snapshot.json")
```
nhinsch, line 17: I wonder if there is a way to refactor this code so that when we add a new stack, we only have to add it once, and the file names are automatically generated?

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,89 @@
...
+yarn
+yarn build
+
+cd $integration_tests_dir
+RAW_CFN_TEMPLATE="cdk.out/ExampleDatadogStack.template.json"
+for ((i=0; i < ${#STACK_CONFIGS[@]}; i++)); do
+    tsc --project tsconfig.json
+    cdk synth --app testlib/integration_tests/stacks/${STACK_CONFIGS[i]} --json
+
+    # Normalize LambdaVersion ID's
```
nhinsch, line 45: Nit: `ID's` -> `IDs`

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,89 @@
...
+
+    cp ${RAW_CFN_TEMPLATE} ${TEST_SNAPSHOTS[i]}
+    echo "===================================="
+    if [ "$UPDATE_SNAPSHOTS" = "true" ]; then
+        echo "Overriding ${CORRECT_SNAPSHOTS[i]}"
+        cp ${TEST_SNAPSHOTS[i]} ${CORRECT_SNAPSHOTS[i]}
+    fi
+
+    echo "Performing diff of ${TEST_SNAPSHOTS[i]} against ${CORRECT_SNAPSHOTS[i]}"
+    set +e # Dont exit right away if there is a diff in snapshots
```
nhinsch, line 76: Nit: `Dont` -> `Don't`

nhinsch:
Can we move the snapshots to a separate directory called `snapshots` within the integration tests directory? I know the Serverless plugin tests don't do that, but some of our other integration tests do, and it's nice to keep the snapshots separated from the rest of the tests.

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,89 @@
...
+# To regenerate snapshots:
+#   UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests.sh
+
+set -e
+
+# To add new tests create a new ts file in the 'integration_tests/lambda' directory, append it as a js file to the STACK_CONFIGS array
+# as well as creating a name for the snapshots that will be compared in your test. Add those snapshot names to the TEST_SNAPSHOTS and CORRECT_SNAPSHOTS arrays.
+# Note: Each yml config, test, and correct snapshot file should be at the same index in their own array. e.g. All the files for the forwarder test are at index 0.
+#       In order for this script to work correctly these arrays should have the same amount of elements.
+STACK_CONFIGS=("lambda-stack.js" "lambda-nodejs-function-stack.js")
```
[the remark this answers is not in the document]
ajsaavedra: I see, would using something like `lambda-function-stack` help differentiate these? You're right they all involve lambdas and the plain `lambda.Function` is the "conventional" way of instantiating one.

scripts/run_integration_tests.sh
ajsaavedra: What if we added two directories `correct_snapshots` and `test_snapshots` and used the for loop to iterate over all three directories (including `stacks`)? Since the files are already sorted by name, they should be run in the correct order.

ajsaavedra:
> I noticed in the test output (in GitHub but I assume locally as well) that we log the entire CloudFormation stack when running the test, even if it's successful. This isn't going to break anything but it could be annoying, especially if we start to add lots of different snapshot tests. Is it possible to make it so that we don't log the full stack?

Yes, there is the flag `--quiet` that suppresses the output to stdout https://www.npmjs.com/package/aws-cdk#user-content-commands. I added that in!

scripts/run_integration_tests.sh
nhinsch: I think you should go with whatever you think the best developer experience will be. You could also have the list of stacks in this file, but then use the stack names to know the names of the test snapshot and correct snapshot files (because they all follow the same pattern).

scripts/run_integration_tests.sh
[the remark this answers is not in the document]
nhinsch: Yeah, I think `lambda-function-stack` is better.

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,89 @@
...
+    set +e # Don't exit right away if there is a diff in snapshots
+    diff ${TEST_SNAPSHOTS[i]} ${CORRECT_SNAPSHOTS[i]}
+    return_code=$?
+    set -e
+    if [ $return_code -eq 0 ]; then
+        echo "SUCCESS: There were no differences between the ${TEST_SNAPSHOTS[i]} and ${CORRECT_SNAPSHOTS[i]}"
+    else
+        echo "FAILURE: There were differences between the ${TEST_SNAPSHOTS[i]} and ${CORRECT_SNAPSHOTS[i]}. Review the diff output above."
+        echo "If you expected the ${TEST_SNAPSHOTS[i]} to be different generate new snapshots by running this command from a development branch on your local repository: 'UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests.sh'"
+        exit 1
```
nhinsch, line 85: It looks like for each stack, we check for a diff and print a success/failure message. If that snapshot failed, we fail the whole integration test suite and exit.

The problem is that if we have, say, six snapshots, and the first one fails, the entire test suite will exit immediately and we won't run the other five snapshot tests.

Can we make it so that all snapshot tests always run, and then at the end we see a summary of which ones passed and which ones failed? (Similar to how unit tests work)

integration_tests/stacks/lambda-stack.ts
```diff
@@ -0,0 +1,40 @@
...
+      apiKey: "1234",
+      site: "datadoghq.com",
+    });
+    datadogCDK.addLambdaFunctions([lambdaFunction]);
+  }
+}
+
+const app = new cdk.App();
+const env = { account: "601427279990", region: "sa-east-1" };
+const stack = new ExampleStack(app, "ExampleDatadogStack", { env: env });
```
nhinsch, line 38: It looks like we call both stacks "ExampleDatadogStack" even though the file names are different. Let's update the stack names to match the file names.

scripts/run_integration_tests.sh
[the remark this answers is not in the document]
nhinsch: It calls attention to the fact that it uses the `Lambda.Function` construct as opposed to the `NodejsFunction` construct

scripts/run_integration_tests.sh
ajsaavedra: Thanks for the feedback! I went with the latter option and just used the stack names to generate those snapshots instead of having to list them. Now that I think about it, I'm not sure that a list for both was ever required!

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,109 @@
...
+# Login to ECR. This is a required step in order to pull a public Docker image for the PythonFunction construct
+aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws
+
+yarn
+yarn build
+
+cd $integration_tests_dir
+
+OUTPUT_ARRAY+=("====================================")
+allTestsPassed=true
```
nhinsch, line 39: By convention, variables in Bash scripts are `ALL_CAPS_AND_SNAKE_CASED`. There are a bunch of variables in this file with different capitalization schemes, let's update them all to be consistent.

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,109 @@
...
+compose_output() {
+     if [ $1 -eq 0 ]; then
+        OUTPUT_ARRAY+=("SUCCESS: There were no differences between the $2 and $3")
+    else
+        allTestsPassed=false
+        OUTPUT_ARRAY+=("FAILURE: There were differences between the $2 and $3. Review the diff output above.")
+        OUTPUT_ARRAY+=("If you expected the $2 to be different generate new snapshots by running this command from a development branch on your local repository: 'UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests.sh'")
+    fi
+    OUTPUT_ARRAY+=("====================================")
+}
```
nhinsch, line 49: I think this is a good improvement, right now the output looks like this:

```
====================================
  asset-output/index.js  249b 
Performing diff of snapshots/test-lambda-function-stack-snapshot.json against snapshots/correct-lambda-function-stack-snapshot.json

SUCCESS: There were no differences between the snapshots/test-lambda-function-stack-snapshot.json and snapshots/correct-lambda-function-stack-snapshot.json
====================================
Performing diff of snapshots/test-lambda-nodejs-function-stack-snapshot.json against snapshots/correct-lambda-nodejs-function-stack-snapshot.json
SUCCESS: There were no differences between the snapshots/test-lambda-nodejs-function-stack-snapshot.json and snapshots/correct-lambda-nodejs-function-stack-snapshot.json
====================================
```

I wonder if we can improve this further and make it look more like the output of our unit tests:

```
PASS: test-lambda-function-stack
PASS: test-lambda-nodejs-function-stack

SUCCESS: 2 out of 2 snapshot tests passed
```

Also, maybe we can log the test results individually as they are completed, then log the overall result at the end? Rather than adding to the output and printing the entire output at the end.

nhinsch: A couple of reasons for this:
- More readable output = faster debugging
- Logging the results live = easier to see the progress of the tests while they run

nhinsch (approved):

PR #72, opened 2021-11-01
Add support for Secrets Manager

duncanista:
### What does this PR do?

Adds support for `DD_API_KEY_SECRET_ARN` for rotating passwords using AWS secrets manager. 
Refer to [SLS-1650](https://datadoghq.atlassian.net/browse/SLS-1650).

### Motivation

Allow users to use their API KEY stored in Secrets Manager.
Also: [Datadog/datadog-cdk-construcs#50](https://github.com/DataDog/datadog-cdk-constructs/pull/50)

### Testing Guidelines

Added unit tests to check multiple keys and the usage of secret ARN in supported cases.

### Additional Notes

N/A

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


src/transport.ts
```diff
@@ -69,6 +74,15 @@ export class Transport {
       if (this.apiKey !== undefined) {
         lam.addEnvironment(API_KEY_ENV_VAR, this.apiKey);
       }
+      const isNode = runtimeLookup[lam.runtime.name] === RuntimeType.NODE
```
nhinsch, line 77: It looks like if you are using synchronous metrics with Node, we will accept the `DD_API_KEY_SECRET_ARN` environment variable in the `validateProps()` function, but then ignore it here. Also, we log a debug log line, which is only visible when debug logs are turned on, but otherwise there is no indication anything is wrong.

What do you want the user experience to be here?

test/datadog.spec.ts
```diff
@@ -254,3 +231,42 @@ describe("addCdkConstructVersionTag", () => {
...
+  it("throws error if both API key secret ARN and KMS API key are defined", () => {
+    expect(() => {
+      checkForMultipleApiKeys({
+        apiKey: "1234",
+        apiKeySecretArn: "some-resource:from:aws:secrets-manager:arn",
+        apiKmsKey: "5678",
+      })
+    }).toThrowError("`apiKey`, `apiKmsKey`, and `apiKeySecretArn` should not be set at the same time.");
+  });
+})
```
nhinsch, line 272: Great test cases 👍

src/transport.ts
duncanista: The main problem is that in `validateProps()` I don't have the runtime of the function, therefore I am not able to check if it will be possible to proceed. Then, I do it in the `transport.ts`. Will try to think of a way to let the user know we are ignoring the secret ARN.

duncanista: Ok, so as suggested, I will change it to an error, since we want the user to know that if he sets the secret ARN and is using sync metrics and a node runtime, it will not work.

README.md
```diff
@@ -72,11 +73,12 @@ _Note_: The descriptions use the npm package parameters, but they also apply to
...
+| `flushMetricsToLogs` | `flush_metrics_to_logs` | Send custom metrics using CloudWatch logs with the Datadog Forwarder Lambda function (recommended). Defaults to `true` . If you disable this parameter, it's required to set `apiKey` (or if encrypted, `apiKMSKey` or `apiKeySecretArn`). |
 | `site` | `site` | Set which Datadog site to send data. This is only used when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com`, `us5.datadoghq.com`, and `ddog-gov.com`. The default is `datadoghq.com`. |
 | `apiKey` | `api_key` | Datadog API Key, only needed when `flushMetricsToLogs` is `false` or `extensionLayerVersion` is set. For more information about getting a Datadog API key, see the [API key documentation][8]. |
+| `apiKeySecretArn` | `api_key_secret_arn` | The ARN of the secret storing the Datadog API key in AWS Secrets Manager. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false` or `extensionLayer` is set. Remember to add the `secretsmanager:GetSecretValue` permission to the Lambda execution role. |
```
nhinsch, line 81: 👌

nhinsch (approved):
🤩

PR #93, opened 2022-02-25
Add source code integration documentation

IvanTopolcic:
This PR adds docs for the source code integration feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,16 @@ datadog.addLambdaFunctions([<LAMBDA_FUNCTIONS>])
 datadog.addForwarderToNonLambdaLogGroups([<LOG_GROUPS>])
 ```
 
+Optionally, if you'd like to enable [source code integration](https://docs.datadoghq.com/integrations/guide/source-code-integration/) (Typescript only), you should add the following:
+
+```typescript
+// Remember to add @datadog/datadog-ci as a dependency
+const datadogCi = require("@datadog/datadog-ci");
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,6 +85,17 @@ datadog.addLambdaFunctions([<LAMBDA_FUNCTIONS>])
 datadog.addForwarderToNonLambdaLogGroups([<LOG_GROUPS>])
 ```
 
+Optionally, if you'd like to enable [source code integration](https://docs.datadoghq.com/integrations/guide/source-code-integration/) (Typescript only), you should add the following:
+
+```typescript
+// Remember to add @datadog/datadog-ci as a dependency
+const datadogCi = require("@datadog/datadog-ci");
+// You may need to slightly refactor your code here and use a top-level async function
```
agocs, line 93: You might want to point to an example. I imagine there's one in the source code integration doc?

IvanTopolcic: I don't believe there's one in the doc. I can go ahead and add one, though.

IvanTopolcic: To clarify -- it's not in the docs because this is really a limitation of the CDK, not of the source code integration feature.

IvanTopolcic: @agocs I've gone ahead and _attempted_ to make it a bit more clear. It's difficult to show a complete example because it requires a change to the CDK constructor. Unfortunately we couldn't add this feature into the CDK because they don't support async operations so it requires a little work from the user to add this feature.

agocs (approved):

PR #102, opened 2022-04-25
Make deployment message clearer

IvanTopolcic:
Makes the deployment message more clear to account for the case where we're deploying v1 and then immediately after, v2.


lizapressman (approved):
LGTM

Issue #124, opened 2022-09-23
Forwarder Subscription Filters Should Not Create Lambda Permissions


jglien:
The current implementation below uses the [default behavior of creating a Lambda permission per log group](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs_destinations.LambdaDestinationOptions.html#addpermissions) on the target lambda (the forwarder) which causes a deployment failure when the number of lambdas/log groups is large enough. The specific error will be:

`The final policy size ({{ size }}) is bigger than the limit (20480)`

https://github.com/DataDog/datadog-cdk-constructs/blob/ebefde300e4c732d8a41545dcfda07bdbe84c9f3/v2/src/forwarder.ts#L28

https://github.com/DataDog/datadog-cdk-constructs/blob/ebefde300e4c732d8a41545dcfda07bdbe84c9f3/v2/src/forwarder.ts#L41

Since the forwarder already has the permission allowing all log groups in the account and region trigger it, the solution is simply to pass `{ addPermissions: false, }`, or make this configurable.

astuyve:
Hi, thanks @jglien! We've triaged this issue and assigned a reviewer who should be able to help you out soon.

In the meantime, I'd suggest using the Datadog Lambda Extension instead of the forwarder. The Extension is our default and recommended integration for Lambda, and will also get you around this issue.

You can do this easily by specifying `extensionLayerVersion` like this:
```js
    const datadog = new Datadog(this, "Datadog", {
      nodeLayerVersion: 83,
      extensionLayerVersion: 29,
      apiKey: '<your key>',
    })
    datadog.addLambdaFunctions([handler])
```

jglien:
@astuyve We have been using the Lambda extension for a while but are switching to the forwarder for two reasons: to improve our lambda performance by offloading the logging overhead and removing a layer, and to allow us to have other log groups send their logs the same way the lambdas do, rather than having multiple integrations to maintain.
Why is the extension recommended over the log group forwarder integration? I don't see that discussed in the docs.

astuyve:
@jglien Interesting! Do you find yourself hitting the maximum 5 layers on many of your functions?

We recommend the Extension over the Forwarder for use with Lambda, and the reasoning is laid out in the [docs](https://docs.datadoghq.com/serverless/guide/extension_motivation/).

Please do let me know if you have further questions. Thanks!

jglien:
[...]

PR #125, opened 2022-09-25
Default to Not Create Forwarder Permissions

jglien:
### What does this PR do?

See #124 

The default behavior of the `LambdaDestination` AWS Log Group CDK construct is to [create a permission on the destination lambda](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs_destinations.LambdaDestinationOptions.html#addpermissions). While this may be desirable in many cases, the Datadog Forwarder already has permission set to allow log groups to forward their logs. Therefore, this default behavior is entirely unnecessary and problematic.

### Motivation

Stacks with large number of Lambdas, and therefore log groups, get resource quota exceeded errors trying to deploy with a Datadog Forwarder. If a stack is split up so there are fewer resources, then a new error shows up: `The final policy size [for the Datadog Forwarder] ({{ size }}) is bigger than the limit (20480)`. This means that this library cannot be used as-is with the Datadog forwarder and anything but a small number of log groups.

### Testing Guidelines

The unit tests won't create a permission resource in the test, so there's nothing to write there, aside from adding the new parameter. Manually testing using the sample, you can observe Lambda policies being created for every Lambda in the template when set to `true`, and they go away when `false`.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ const datadog = new Datadog(this, "Datadog", {
   addLayers: <BOOLEAN>,
   extensionLayerVersion: "<EXTENSION_VERSION>",
   forwarderArn: "<FORWARDER_ARN>",
+  createForwarderPermissions: <BOOLEAN>,
   flushMetricsToLogs: <BOOLEAN>,
   site: "<SITE>",
   apiKey: "{Datadog_API_Key}",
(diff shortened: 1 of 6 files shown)


jglien:
@lizapressman @astuyve @cgatt Not sure if you all have seen this yet.

IvanTopolcic (approved):

PR #151, opened 2023-02-10
Enable source code integration within the CDK itself + add python support

thedavl:
### What does this PR do?

uses `execSync` from node's `child_processes` module to synchronously call `git rev-parse HEAD` and `git config --get remote.origin.url` to get the commit sha + repo url.
- Note: github repo urls are stripped of any sensitive information + are formatted. This code was copied from the datadog-ci

the `git.commit.sha` and `git.repository_url` tags are then added by default to the lambda functions.

if either of these commands fail, a warning will be printed, but deployment will continue.

users can turn off source code integration by setting the `sourceCodeIntegration` config value to false.

If users don't have the datadog github integration added, they will still need to import datadog-ci themselves and upload metadata. this will be documented in a future docs PR
- unused parameters in `addGitCommitMetadata` are kept for backwards compatibility, so previous ways of enabling source code integration don't break

### Testing Guidelines

updated unit tests to test both sourceCodeIntegration set to true and false. Manually tested by compiling the js and py packages, using them to deploy v1 & v2 python & node CDK stacks.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/common/env.ts
+++ b/common/env.ts
@@ -26,7 +26,7 @@ const URL = require("url").URL;
 
 export function setGitEnvironmentVariables(lambdas: any[]) {
   log.debug("Adding source code integration...");
-  const [hash, gitRepoUrl] = getGitData();
+  const { hash, gitRepoUrl } = getGitData();
 
   if (hash == "" || gitRepoUrl == "") return;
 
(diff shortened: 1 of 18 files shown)


common/env.ts
```diff
@@ -20,20 +20,72 @@ export const DD_SERVICE_ENV_VAR = "DD_SERVICE";
...
 export const DD_TAGS = "DD_TAGS";
 
-export function setGitCommitEnvironmentVariables(lambdas: any[], hash: string, gitRepoUrl: string | undefined) {
+const execSync = require("child_process").execSync;
+
+const URL = require("url").URL;
+
+export function setGitEnvironmentVariables(lambdas: any[]) {
+  log.debug("Adding source code integration...");
+  const [hash, gitRepoUrl] = getGitData();
```
IvanTopolcic, line 29: nit: I'm usually against returning lists because it doesn't tell us anything about the meaning of the return values. Might be better to return an object, but it probably doesn't matter for this use case.

IvanTopolcic (approved):
LGTM, nice work!

common/env.ts
thedavl: addressed

Issue #174, opened 2023-04-27
Automatically grantRead for `apiKeySecretArn`


blimmer:
Currently, the documentation mentions that you have to remember to manually grant `GetSecretValue` when passing the API key as an ARN: https://github.com/DataDog/datadog-cdk-constructs/blob/8ffefd8dd5432805de312ade0f60d1015682b123/README.md#L175

It would be nice if this construct did it for you.

Instead of passing a secret arn string:

```ts
new Datadog(scope, 'Datadog', {
  apiKeySecretArn: 'arn:aws:secretsmanager:us-west-2:12345678910:secret:DATADOG_API_KEY'
});
```

If the user could pass an `ISecret`:

```ts
const secret = Secret.fromSecretPartialArn(this, 'DatadogApiKeySecret', 'arn:aws:secretsmanager:us-west-2:12345678910:secret:DATADOG_API_KEY');
new Datadog(scope, 'Datadog', {
  apiKeySecret: secret
});
```

Then, in the construct, you could `grantRead` for us:

```ts
apiKeySecret.grantRead(lambda)
```

sfirrin:
Hi @blimmer thanks for this suggestion. We're going to work on implementing this and I will leave this ticket open for tracking

blimmer:
Btw, this is made worse by the fact that the extension doesn't tell you clearly that you're missing this permission: https://github.com/DataDog/datadog-agent/issues/16927

thedavl:
Hi @blimmer, we've implemented this feature and it has been included in the most recent `v2-1.7.0` release ([pypi](https://pypi.org/project/datadog-cdk-constructs-v2/), [npm](https://www.npmjs.com/package/datadog-cdk-constructs-v2)).

We're working on getting the associated docs changes merged, but in the meantime here's an example of how to use it (just like what you suggested):

```
const { Secret } = require('aws-cdk-lib/aws-secretsmanager');
const secret = Secret.fromSecretPartialArn(this, 'DatadogApiKeySecret' 'arn:aws:secretsmanager:us-west-1:123:secret:DATADOG_API_KEY');

const datadog = new Datadog(scope, 'Datadog', {
  ...
  apiKeySecret: secret
  ...
});
```

Thanks for your suggestion, please let me know if you've got any questions!

PR #181, opened 2023-05-18
Fix: auto grant secret read permissions python package bug

thedavl:
### What does this PR do?

[This PR](https://github.com/DataDog/datadog-cdk-constructs/issues/174) worked as a JS package, but produced a broken Python package. This was not caught in testing.

The issue stems from a bug with typescript intersection types compiled into python, [see here](https://github.com/DataDog/datadog-cdk-constructs/blob/2343119f325c942768291f1f05bbd292d5dc0c0a/v2/src/datadog.ts#L35) for the exact line.

This PR fixes this

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/v2/src/datadog.ts
+++ b/v2/src/datadog.ts
@@ -23,7 +23,7 @@ import {
   applyEnvVariables,
   validateProps,
   TagKeys,
-  IDatadogProps,
+  DatadogPropsV2,
   DatadogStrictProps,
   handleSettingPropDefaults,
   setGitEnvironmentVariables,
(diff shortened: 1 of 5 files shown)


v2/src/datadog.ts
```diff
@@ -23,7 +23,7 @@ import {
   applyEnvVariables,
   validateProps,
   TagKeys,
-  IDatadogProps,
+  DatadogPropsV2,
```
thedavl, line 26: we need to include ISecret (specific to CDK v2) in our props interface. Using type intersection (`&`) doesn't work with JSII (the bug introduced that this PR is fixing). Therefore we need to create a v2 specific interface that includes ISecret.

v2/src/datadog.ts
```diff
@@ -32,28 +32,27 @@ import {
...
     if (process.env.DD_CONSTRUCT_DEBUG_LOGS?.toLowerCase() == "true") {
       log.setLevel("debug");
     }
     super(scope, id);
     this.scope = scope;
     this.props = props;
+    let apiKeySecretArn = this.props.apiKeySecretArn;
     if (this.props.apiKeySecret !== undefined) {
-      this.props.apiKeySecretArn = this.props.apiKeySecret.secretArn;
+      apiKeySecretArn = this.props.apiKeySecret.secretArn;
```
thedavl, line 48: declaring `DatadogPropsV2 ` as a behavioral interface is preferred as we can have non-readonly values, so that we can directly mutate apiKeySecretArn.

This isn't possible because it changes the way the datadog construct is declared, thus making it non-backwards compatable. See below:

before:
```
datadog = Datadog(
            self,
            "Datadog",
            extension_layer_version=43,
            python_layer_version=72,
            site="datadoghq.com",
            api_key_secret=secret,
        )
```

non-backwards compatible change introduced IF we use behavioral interfaces:
```
datadog = Datadog(
            self,
            "Datadog",
            {
                extension_layer_version=43,
                python_layer_version=72,
                site="datadoghq.com",
                api_key_secret=secret,
            }
        )
```

therefore we can't use behavioral interfaces, and have to live with passing a apiKeySecretArn override flag as a parameter to `validateProps `

sfirrin (approved):

PR #183, opened 2023-06-02
auto grant red secret permissions from a secret arn

thedavl:
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/v2/src/datadog.ts
+++ b/v2/src/datadog.ts
@@ -11,7 +11,7 @@ import { Tags } from "aws-cdk-lib";
 import * as lambda from "aws-cdk-lib/aws-lambda";
 import * as lambdaNodejs from "aws-cdk-lib/aws-lambda-nodejs";
 import * as logs from "aws-cdk-lib/aws-logs";
-import { ISecret } from "aws-cdk-lib/aws-secretsmanager";
+import { ISecret, Secret } from "aws-cdk-lib/aws-secretsmanager";
 import { Construct } from "constructs";
 import log from "loglevel";
 import { Transport } from "./common/transport";
@@ -60,6 +60,7 @@ export class Datadog extends Construct {
 
   public addLambdaFunctions(
     lambdaFunctions: (lambda.Function | lambdaNodejs.NodejsFunction | lambdaPython.PythonFunction)[],
+    construct?: Construct,
   ) {
     // baseProps contains all properties set by the user, with default values for properties
     // defined in DefaultDatadogProps (if not set by user)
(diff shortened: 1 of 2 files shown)


v2/src/datadog.ts
```diff
@@ -185,3 +189,10 @@ function grantReadLambdas(secret: ISecret, lambdaFunctions: lambda.Function[]) {
     secret.grantRead(functionName);
   });
 }
+
+function grantReadLambdasFromSecretArn(construct: Construct, arn: string, lambdaFunctions: lambda.Function[]) {
+  const secret = Secret.fromSecretPartialArn(construct, "DatadogApiKeySecret", arn);
+  lambdaFunctions.forEach((functionName) => {
+    secret.grantRead(functionName);
```
astuyve, line 196: Users deploying these applications may not have permissions to grant read access to keys in SSM or KMS - is there a way we can make this optional?

thedavl: if users don't pass the construct to `addLambdaFunctions` we won't try to grant anything. We could document this explicitly, only pass in the construct if you want us to try auto-granting secret read permissions?

thedavl: the optional construct param is only used for this feature

astuyve: We could, but we may need the construct for other work in the near future.

Can we default this to true but allow users to disable this behavior if needed?

It's (occasionally) common to see devops teams own secrets management and have elevated permissions to create or grant read permissions to individual applications, so that individual developer teams can't read arbitrary keys out of secrets storage.

thedavl: Gotcha, makes sense. I'll add a config param for disabling it!

v2/src/datadog.ts
```diff
@@ -68,6 +69,13 @@ export class Datadog extends Construct {
     if (this.props !== undefined && lambdaFunctions.length > 0) {
       if (this.props.apiKeySecret !== undefined) {
         grantReadLambdas(this.props.apiKeySecret, lambdaFunctions);
+      } else if (
+        this.props.apiKeySecretArn !== undefined &&
+        construct !== undefined &&
+        baseProps.grantSecretReadAccess
+      ) {
+        log.debug("Granting read access to the provided Secret ARN for all your lambda functions.");
```
astuyve, line 77: Looks good!

astuyve (approved):
Great work on this one David, very nicely done!

PR #192, opened 2023-06-27
Allow bypassing site check

nhulston:
### What does this PR do?

Allow bypassing the site-check when the environment variable `DD_CDK_BYPASS_SITE_VALIDATION` is set to "true"

### Motivation

Datadog developers need a way to bypass the site-check to manually test the CDK tool on staging or any other custom site that is not in `siteList`.

### Testing Guidelines

I wrote new tests for v1 and v2 to check that the site check passes when using an invalid site and `DD_CDK_BYPASS_SITE_VALIDATION` is set.

### Additional Notes

This is not intended to be used by users and only for developing and testing purposes.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/common/datadogSharedLogic.ts
+++ b/common/datadogSharedLogic.ts
@@ -26,7 +26,7 @@ export function validateProps(props: DatadogProps, apiKeyArnOverride = false) {
     props.site !== undefined &&
     !siteList.includes(props.site.toLowerCase()) &&
     !(props.site.startsWith("${Token[") && props.site.endsWith("]}")) &&
-    process.env.DD_CDK_BYPASS_SITE_VALIDATION !== "true"
+    !process.env.DD_CDK_BYPASS_SITE_VALIDATION
   ) {
     throw new Error(
       "Warning: Invalid site URL. Must be either datadoghq.com, datadoghq.eu, us3.datadoghq.com, us5.datadoghq.com, ap1.datadoghq.com, or ddog-gov.com.",
--- a/v1/test/datadog.spec.ts
+++ b/v1/test/datadog.spec.ts
@@ -79,7 +79,7 @@ describe("validateProps", () => {
     }
     expect(threwError).toBe(false);
     expect(thrownError?.message).toEqual(undefined);
-    process.env.DD_CDK_BYPASS_SITE_VALIDATION = "undefined";
+    process.env.DD_CDK_BYPASS_SITE_VALIDATION = undefined;
   });
 
   it("doesn't throw an error when the site is set as a cdk token", () => {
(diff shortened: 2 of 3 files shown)


common/datadogSharedLogic.ts
```diff
@@ -25,7 +25,8 @@ export function validateProps(props: DatadogProps, apiKeyArnOverride = false) {
   if (
     props.site !== undefined &&
     !siteList.includes(props.site.toLowerCase()) &&
-    !(props.site.startsWith("${Token[") && props.site.endsWith("]}"))
+    !(props.site.startsWith("${Token[") && props.site.endsWith("]}")) &&
+    process.env.DD_CDK_BYPASS_SITE_VALIDATION !== "true"
```
astuyve, line 29: So this will still throw an error if the user specifies `DD_CDK_BYPASS_SITE_VALIDATION=True` or `DD_CDK_BYPASS_SITE_VALIDATION=TRUE`.

I think we want to lowercase that?

Worth considering this solution versus checking for this to be `false`, or even checking for the presence of the env var at all?

(to be clear I think we should lowercase and call it good, I'm just laying out the alternative as a thought exercise)

astuyve (approved):
LGTM, nice work on the test cases. One note, I think we should lowercase the env check.

common/datadogSharedLogic.ts
nhulston: Good catch! I will lowercase the env variable first.

Note that if the environment variable is not set at all, set to `false`, or set to anything but `true`, the site check will still throw an error.
If we instead checked `process.env. DD_CDK_BYPASS_SITE_VALIDATION === "false"`, then the site check would be bypassed unless they set DD_CDK_BYPASS_SITE_VALIDATION to false.

astuyve: Exactly - it's just something to think about. Would anyone actually set `DD_CDK_BYPASS_SITE_VALIDATION` to false? Probably not. They'd only set the env var if they actually want to bypass the site check (eg: the intent of the config var is declared by the presence of the env var).

We see this in other cases too like `DD_TRACING_ENABLED`. But compare that to an env var that is not a boolean, like `DD_SITE` where we have different assumptions. We may consider checking just for `presence` here versus `value`.

nhulston: Ah, I see your point. If we want to check for only presence, I can do something like:
`&& !process.env.DD_CDK_BYPASS_SITE_VALIDATION`.

Which option do you think would be better? I'm open to doing both

astuyve: Ohhh I see. In this specific case I think let's be consistent with `datadog-ci` then.

Issue #185, opened 2023-06-09
Can't add env and service values to a Docker Lambda


icaspi:
## Expected Behavior
When setting up the datadog construct for a lambda the `env` and `service` are added with the extension layer and add environment variables to the lambda

## Actual Behavior
Adding extension layer is being reject because it is a packaged lambda. Deploying without the extension is working as expected but don't create the environment variables 

## Steps to Reproduce the Problem

  1. Create a CDK script
  2. Add to it function from a docker file
  3. Create a datadog component and include `extensionLayerVersion`  key
  

## Specifications

  - Datadog Lambda Layer version: 43 
  - Node version: 
  
## Stacktrace
  
  ```
 "Please don't provide Handler or Runtime or Layer when the intended function PackageType is Image. (Service: Lambda, Status Code: 400, Request ID: 0db9418c-418f-4c5b-b6a3-f89ec1d9f873)" (RequestToken: 75a65225-d578-b731-7070-0558c4690f9a, HandlerErrorCode: InvalidRequest)
    at FullCloudFormationDeployment.monitorDeployment (/usr/local/lib/node_modules/aws-cdk/lib/index.js:380:10236)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async deployStack2 (/usr/local/lib/node_modules/aws-cdk/lib/index.js:383:145458)
    at async /usr/local/lib/node_modules/aws-cdk/lib/index.js:383:128776
    at async run (/usr/local/lib/node_modules/aws-cdk/lib/index.js:383:126782)
  ```

sfirrin:
Hi @icaspi, thanks for opening this

We will look into a fix and update you here

sfirrin:
To update @icaspi, I have just been able to reproduce and am currently working on a fix

sfirrin:
Hi again @icaspi, we just released https://github.com/DataDog/datadog-cdk-constructs/releases/tag/v2-1.7.4 with the fix and in 1.7.4 I no longer get the error in my testing Docker CDK app. Thanks for opening the issue and please get in touch again if you run into any other problem

PR #193, opened 2023-07-03
Fix error when instrumenting Lambda container images 

sfirrin:
### What does this PR do?

- Fixes the error brought up in https://github.com/DataDog/datadog-cdk-constructs/issues/185 by preventing the extension layer from being added to a Lambda container image
- Adds tests for the container image case

### Motivation
https://github.com/DataDog/datadog-cdk-constructs/issues/185

### Testing Guidelines
Deployed with example CDK Lambda container image project with the error no longer happening

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/v2/src/layer.ts
+++ b/v2/src/layer.ts
@@ -42,8 +42,6 @@ export function applyLayers(
       lam.architecture.dockerPlatform === Architecture.ARM_64.dockerPlatform;
     const isNode = lambdaRuntimeType === RuntimeType.NODE;
 
-    // console.log({ lam });
-
     if (lambdaRuntimeType === undefined || lambdaRuntimeType === RuntimeType.UNSUPPORTED) {
       log.debug(`Unsupported runtime: ${runtime}`);
       return;
--- a/v2/src/layer.ts
+++ b/v2/src/layer.ts
@@ -41,7 +41,10 @@ export function applyLayers(
       lam.architecture?.dockerPlatform !== undefined &&
       lam.architecture.dockerPlatform === Architecture.ARM_64.dockerPlatform;
     const isNode = lambdaRuntimeType === RuntimeType.NODE;
-    if (lambdaRuntimeType === RuntimeType.UNSUPPORTED) {
+
+    // console.log({ lam });
+
+    if (lambdaRuntimeType === undefined || lambdaRuntimeType === RuntimeType.UNSUPPORTED) {
       log.debug(`Unsupported runtime: ${runtime}`);
       return;
     }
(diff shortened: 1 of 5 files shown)


v2/src/layer.ts
```diff
@@ -41,7 +41,8 @@ export function applyLayers(
       lam.architecture?.dockerPlatform !== undefined &&
       lam.architecture.dockerPlatform === Architecture.ARM_64.dockerPlatform;
     const isNode = lambdaRuntimeType === RuntimeType.NODE;
-    if (lambdaRuntimeType === RuntimeType.UNSUPPORTED) {
+
+    if (lambdaRuntimeType === undefined || lambdaRuntimeType === RuntimeType.UNSUPPORTED) {
       log.debug(`Unsupported runtime: ${runtime}`);
```
IvanTopolcic, line 46: Just checking but what's the value of `runtime` if `lambdaRuntimeType` is undefined? Is that just going to print undefined?

v2/test/layer.spec.ts
```diff
@@ -236,6 +236,62 @@ describe("applyLayers", () => {
...
+  it("doesn't add layer to container image Lambda without extension or layer versions", () => {
+    const app = new App();
+    const stack = new Stack(app, "stack", {
+      env: {
+        region: "us-west-2",
+      },
+    });
+    const hello = new lambda.DockerImageFunction(stack, "HelloHandler", {
+      functionName: "container-lambda",
+      code: lambda.DockerImageCode.fromImageAsset("./test/assets"),
```
IvanTopolcic, line 248: nit: does this default to `Dockerfile` if a dir is specified? It might be a little more clear to explicitly specify the Dockerfile?

IvanTopolcic (approved):

v2/src/layer.ts
sfirrin: Yeah for container images it's also undefined

v2/test/layer.spec.ts
sfirrin: Yep it's the dir of the Dockerfile. I just tried out specifying the Dockerfile itself and got this error, it has to be the dir: `Cannot find file at /Users/stephen.firrincieli/dev/datadog-cdk-constructs/v2/test/assets/Dockerfile/Dockerfile`

v2/src/layer.ts
IvanTopolcic: LGTM

v2/test/layer.spec.ts
IvanTopolcic: ah okay interesting. I think that's okay for now. maybe if we decide to add future Dockerfiles at some point we can revisit.

PR #207, opened 2023-08-15
adds support for singleton functions

duncanpharvey:
### What does this PR do?

Adds support for Singleton functions.

### Motivation

https://datadoghq.atlassian.net/browse/SVLS-3474

### Testing Guidelines

* Added unit test to verify `addLambdaFunctions` handles Singleton functions as expected
* Added integration test for adding Datadog extension and layer to Singleton function
  * `./v2/scripts/run_integration_tests.sh` (with `UPDATE_SNAPSHOTS-true` to generate snapshot for singleton function)
* Tested manually in sandbox
  * `npx cdk --app lib/sample/index.js deploy`

### Additional Notes

The Lambda function in the SingletonFunction class is in the `lambdaFunction` property so before making any updates to Singleton functions the Lambda function is extracted from this `lambdaFunction` property.

* Updated `test-utils.findDatadogSubscriptionFilters` to properly handle Singleton functions in unit test
* Added singleton function handling to `datadog.addGitCommitMetadata` since it is also a public method along with `datadog.addLambdaFunctions`

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/v2/src/datadog.ts
+++ b/v2/src/datadog.ts
@@ -203,9 +203,9 @@ function grantReadLambdasFromSecretArn(construct: Construct, arn: string, lambda
 function extractSingletonFunctions(lambdaFunctions: LambdaFunction[]) {
   // extract lambdaFunction property from Singleton Function
   // using any here since lambdaFunction is a private property
-  const extractedLambdaFunctions: lambda.Function[] = lambdaFunctions.map((fn: any) => {
+  const extractedLambdaFunctions: lambda.Function[] = lambdaFunctions.map((fn) => {
     if (fn.hasOwnProperty("lambdaFunction")) {
-      return fn.lambdaFunction;
+      return (fn as any).lambdaFunction;
     }
     return fn;
   });
--- a/v2/src/datadog.ts
+++ b/v2/src/datadog.ts
@@ -202,10 +202,10 @@ function grantReadLambdasFromSecretArn(construct: Construct, arn: string, lambda
 
 function extractSingletonFunctions(lambdaFunctions: LambdaFunction[]) {
   // extract lambdaFunction property from Singleton Function
-  // using any here since lambdaFunction is a private property
+  // using bracket notation here since lambdaFunction is a private property
   const extractedLambdaFunctions: lambda.Function[] = lambdaFunctions.map((fn) => {
     if (fn.hasOwnProperty("lambdaFunction")) {
-      return (fn as any).lambdaFunction;
+      return (fn as lambda.SingletonFunction)["lambdaFunction"]; // eslint-disable-line dot-notation
     }
     return fn;
   });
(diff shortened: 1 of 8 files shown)


v2/src/datadog.ts
```diff
@@ -59,7 +59,12 @@ export class Datadog extends Construct {
   }
 
   public addLambdaFunctions(
-    lambdaFunctions: (lambda.Function | lambdaNodejs.NodejsFunction | lambdaPython.PythonFunction)[],
+    lambdaFunctions: (
+      | lambda.Function
+      | lambdaNodejs.NodejsFunction
+      | lambdaPython.PythonFunction
+      | lambda.SingletonFunction
```
sfirrin, line 66: Can you put this into a named type and then use that type in all the places you're using the union in this PR?

sfirrin (commented):
Nice work, left one comment about creating a named type. Otherwise once the conflicts are fixed this should be good to go

v2/src/datadog.ts
```diff
@@ -201,6 +200,19 @@ function grantReadLambdasFromSecretArn(construct: Construct, arn: string, lambda
   });
 }
 
+function extractSingletonFunctions(lambdaFunctions: LambdaFunction[]) {
+  // extract lambdaFunction property from Singleton Function
+  // using any here since lambdaFunction is a private property
```
sfirrin, line 205: I think you can get around the any cast by using the bracket notation to access the lambdaFunction property https://dev.to/zirkelc/hack-access-private-class-members-in-typescript-41jp

sfirrin (commented):
Thank you for adding the integration test :clap: great work

Just left one comment about types

v2/src/datadog.ts
duncanpharvey: It looks like this `dot-notation` rule treats the bracket notation as an error.

https://github.com/DataDog/datadog-cdk-constructs/blob/0bf6d168ed5aaa6615a78021fc2aa5aa4cf60c32/v2/.eslintrc.json#L197-L199

But based on the article you shared I made a minor update to at least limit the any cast to only Singleton functions instead of all functions. Let me know if that's acceptable

sfirrin: Ah I see, that's trickier than I thought. Since the typechecking can help us avoid runtime errors and the linting is more just for a consistent style, I would do a single-line eslint override here to allow the bracket notation

duncanpharvey: Got it, I changed the type cast to a lambda.SingletonFunction and added the single-line eslint override for the dot-notation rule

The cast is still needed because the lambdaFunction property is only on the lambda.SingletonFunction type and not on the lambda.Function, lambdaNodejs.NodejsFunction, or lambdaPython.PythonFunction types (which are all part of the custom LambdaFunction type).

Screenshot 2023-09-19 at 5 15 13 PM

sfirrin (approved):

v2/src/datadog.ts
duncanpharvey: Refactored to use the type guard as we discussed

PR #236, opened 2023-12-04
feat: support `.NET` runtimes

duncanista:
### What does this PR do?

Allows .NET AWS Lambdas to be instrumented using this CDK construct.

### Motivation

We allow this in other Datadog instrumentation tools.

### Testing Guidelines

Tested manually, added an example for both the `typescript` and `python` stacks under `examples`.
Also added a unit test for the wrapper.

### Additional Notes

I also removed some outdated runtimes from our unit tests.
Didn't create integration tests yet.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


alai97 (approved):

v2/src/layer.ts
```diff
@@ -60,7 +60,7 @@ export function applyLayers(
         errors.push(errorMessage);
         return;
       }
-      lambdaLayerArn = getLambdaLayerArn(region, pythonLayerVersion, runtime, isARM, isNode, accountId);
+      lambdaLayerArn = getLambdaLayerArn(region, pythonLayerVersion, runtime, isARM, false, accountId);
```
alexgallotta, line 63: I noticed that there is a lot of repeated logic here. Can it be deduplicated?

v2/src/redirect.ts
```diff
@@ -29,42 +29,49 @@ import {
...
+    const runtimeType: RuntimeType = runtimeLookup[runtime];
+
+    if (runtimeType === RuntimeType.JAVA || runtimeType === RuntimeType.DOTNET) {
+      l.addEnvironment(AWS_LAMBDA_EXEC_WRAPPER_ENV_VAR, AWS_LAMBDA_EXEC_WRAPPER);
+      continue;
+    }
+
+    const cfnFuntion = l.node.defaultChild as lambda.CfnFunction;
+    if (cfnFuntion === undefined) {
+      log.debug("Unable to get Lambda Function handler");
```
alexgallotta, line 44: should this be a warn rather than debug log?

v2/src/redirect.ts
```diff
@@ -29,42 +29,49 @@ import {
...
+    cfnFuntion.handler = handler;
+  }
 }
 
-function getDDHandler(lambdaRuntime: RuntimeType, addLayers: boolean) {
-  if (lambdaRuntime === undefined || lambdaRuntime === RuntimeType.UNSUPPORTED) {
+function getDDHandler(runtimeType: RuntimeType, addLayers: boolean) {
+  if (runtimeType === undefined || runtimeType === RuntimeType.UNSUPPORTED) {
     log.debug("Unsupported/undefined Lambda runtime");
     return;
```
alexgallotta, line 64: is it better to have return null here? I think the undefined would cause unintended behavior

v2/test/env.spec.ts
```diff
@@ -344,7 +344,7 @@ describe("INJECT_LOG_CONTEXT_ENV_VAR", () => {
       },
     });
     const hello = new lambda.Function(stack, "HelloHandler", {
-      runtime: lambda.Runtime.NODEJS_12_X,
+      runtime: lambda.Runtime.NODEJS_16_X,
```
alexgallotta, line 347: there is 10x18 = 180 loc that can be extracted to a function here. Is it worth extracting it?
```
const app = new App();
    const stack = new Stack(app, "stack", {
      env: {
        region: "us-west-2",
      },
    });
    const hello = new lambda.Function(stack, "HelloHandler", {
      runtime: lambda.Runtime.NODEJS_16_X,
      code: lambda.Code.fromInline("test"),
      handler: "hello.handler",
    });
    ```

**Review by alexgallotta — COMMENTED** [review]
Looks good! But left a few comments (especially the one of undefined vs null)
It would be great to remove deduplicate code where it doesn't affect readibility, especially in tests

File: v2/src/layer.ts (hunk 1 above)
Comment (duncanista) [reply v2/src/layer.ts:63]: Yup, I wanna address it in another PR, too much to refactor if I do it here.

File: v2/src/redirect.ts (hunk 2 above)
Comment (duncanista) [reply v2/src/redirect.ts:44]: Not really, I think Container based Lambdas don't have a handler AFAIK.

File: v2/src/constants.ts [hunk 5]
```diff
@@ -48,6 +49,7 @@ export enum TagKeys {
 }
 
 export const runtimeLookup: { [key: string]: RuntimeType } = {
+  "dotnet6": RuntimeType.DOTNET,
```
astuyve, line 52: dotnet 8 is in a month, worth adding now? https://github.com/aws/aws-lambda-dotnet/issues/1611#issuecomment-1831523596

v2/test/env.spec.ts
duncanista: I'll take a look into it.

duncanista: By definition, we will still be doing the same allocations, due to the state of the stack being changed on every test. I'd rather do this is in another PR where we restructure testing. WDYT?

v2/src/constants.ts
duncanista: I could add it, yet I cannot promise it will work.
On the same not, we'd have to add `Python_3_12` since its supposed to be released this month. I have no issues, for Python, since the definitions are already available in the CDK. Cannot say the same for `DOTNET_8`.

duncanista: As per our conversation, i'm going to close this and marked as resolved, will be addressed in separate PRs.

joeyzhao2018 (approved):
LGTM

PR #288, opened 2024-08-29
chore: rename Lambda related interfaces

lym953:
### What does this PR do?
1. Rename: `DatadogProps` -> `DatadogLambdaProps`
2. Rename: `DefaultDatadogProps` -> `DatadogLambdaDefaultProps`
3. Rename: `DatadogStrictProps` -> `DatadogLambdaStrictProps`
4. Add a type alias `export type DatadogProps = DatadogLambdaProps;` so the renaming won't affect users

### Motivation

To make it cleaner to add `DatadogStepFunction` class later. Details in [[RFC] Changing API for Datadog CDK Construct](https://docs.google.com/document/d/1g3bFuhWUayvuS_aS6kOHe_CGECocd8IhoKmM9s4dPU0)

### Testing Guidelines

`npm test`

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -27,7 +27,7 @@ export enum RuntimeType {
   UNSUPPORTED,
 }
 
-export const DefaultDatadogProps = {
+export const DatadogLambdaDefaultProps = {
   addLayers: true,
   enableDatadogTracing: true,
   enableDatadogASM: false,
--- a/src/datadog.ts
+++ b/src/datadog.ts
@@ -1,3 +1,8 @@
 import { DatadogLambda } from "./datadog-lambda";
 
+/**
+ * For backward compatibility. To be deprecated.
+ * It's recommended to use DatadogLambda for users who want to add Datadog
+ * monitoring for Lambda functions.
+ */
 export class Datadog extends DatadogLambda {}
--- a/src/datadog-lambda.ts
+++ b/src/datadog-lambda.ts
@@ -19,11 +19,11 @@ import {
   addForwarderToLogGroups,
   applyEnvVariables,
   TagKeys,
-  DatadogStrictProps,
+  DatadogLambdaStrictProps,
   setGitEnvironmentVariables,
   setDDEnvVariables,
-  DefaultDatadogProps,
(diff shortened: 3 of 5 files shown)


src/constants.ts
```diff
@@ -27,7 +27,7 @@ export enum RuntimeType {
   UNSUPPORTED,
 }
 
-export const DefaultDatadogProps = {
+export const DefaultDatadogLambdaProps = {
```
purple4reina, line 30: I kinda prefer `DatadogLambdaDefaultProps` but only because it falls in line with the naming of the other props types

lym953: I couldn't find another default prop type. Could you give an example?

purple4reina: Sorry, I think I wasn't clear. My (nitpick) preference is that this prop be renamed `DatadogLambdaDefaultProps` rather than `DefaultDatadogLambdaProps`. The other prop types I was referring to were:

`DatadogLambdaStrictProps`  
`DatadogLambdaProps`

Notice that these both start with `Datadog`, where as the default one doesn't.

src/datadog.ts
```diff
@@ -1,3 +1,7 @@
 import { DatadogLambda } from "./datadog-lambda";
 
+/**
+ * For backward compatibility. It's recommended to use DatadogLambda for
+ * users who want to add Datadog monitoring for Lambda functions.
+ */
```
purple4reina, line 6: Could we include the word "Deprecated" in this statement somewhere? This will tell us and users what to expect when we do our next major version update.

src/constants.ts
lym953: Make sense! Will rename it.

src/datadog.ts
lym953: Sure, will add "To be deprecated.".

purple4reina (approved):

automartin5000:
This is a breaking change
https://github.com/DataDog/datadog-cdk-constructs/issues/294

Issue #294, opened 2024-09-09
Breaking Change in v2-1.16.0


automartin5000:
It looks like y'all just released a major breaking change in v2-1.16.0 to the Datadog API interface by renaming [DatadogProps to DatadogLambdaProps](https://github.com/DataDog/datadog-cdk-constructs/pull/288/files). It seems like that should be reverted or the major version bumped.

pwrmiller:
Confirmed here. In Python, you'll see an error message similar to:

```
ImportError: cannot import name 'DatadogProps' from 'datadog_cdk_constructs_v2'
```

I've added the error message here to assist with visibility, and agree that this represents a major version/ breaking change.

lym953:
Thanks for reporting. It should have been fixed in v1.16.1. Could you retry?

JuliaLid:
I can confirm that updating to v1.16.1 fixed the issue that @pwrmiller reported earlier.

automartin5000:
Confirmed working from our side. Thanks for the quick revert.

PR #295, opened 2024-09-09
fix: make DatadogProps available in Python again

lym953:
### What does this PR do?
Solves a few backward compatibility issues:
- Change `DatadogProps` from a type to an interface, so there will be a class `DatadogProps` in the generated Python package
- Create an interface `DatadogStrictProps` (which has been renamed to `DatadogLambdaStrictProps` in #288), in case it's still being used by any external code
- Create a const `DefaultDatadogProps` (which has been renamed to `DatadogLambdaDefaultProps` in #288), in case it's still being used by any external code

### Motivation
Users reported that the class `DatadogProps` is no longer available in the Python package v2-1.16.0. https://github.com/DataDog/datadog-cdk-constructs/issues/294

#### Steps
1. Make changes on the example Python stack to make it use `DatadogProps` https://github.com/DataDog/datadog-cdk-constructs/commit/480eed099a4836360e72365db6edcd1af1d43d36
2. Deploy the stack
#### Result
Before:
- `cdk deploy` gives an error:
> ImportError: cannot import name 'DatadogProps' from 'datadog_cdk_constructs_v2'

After:
- `cdk deploy` runs successfully

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -52,7 +52,7 @@ export interface DatadogLambdaProps {
  * For backward compatibility. It's recommended to use DatadogLambdaProps for
  * users who want to add Datadog monitoring for Lambda functions.
  */
-export type DatadogProps = DatadogLambdaProps;
+export interface DatadogProps extends DatadogLambdaProps {}
 
 /*
  * Makes fields shared with DatadogLambdaDefaultProps (in constants file) required.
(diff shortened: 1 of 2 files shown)


src/interfaces.ts
```diff
@@ -52,7 +52,7 @@ export interface DatadogLambdaProps {
  * For backward compatibility. It's recommended to use DatadogLambdaProps for
  * users who want to add Datadog monitoring for Lambda functions.
  */
-export type DatadogProps = DatadogLambdaProps;
+export interface DatadogProps extends DatadogLambdaProps {}
```
lym953, line 55: There won't be a class `DatadogProps` in the generated Python package if we use `export type DatadogProps` here. The will be such a Python class if we use `export interface DatadogProps` here.

lym953:
Since #288 is reverted, I'll rewrite the current PR and open another one.

PR #285, opened 2024-08-28
feat: Rename: Datadog -> DatadogLambda

lym953:
### What does this PR do?

1. Rename `Datadog` class as `DatadogLambda`, including:
    1. rename `datadog.ts` as `datadog-lambda.ts`
    2. rename `datadog.spec.ts` as `datadog-lambda.spec.ts` 
3. Make all references of `Datadog` use `DatadogLambda`
4. Create a new `Datadog` class which just wraps (extends) the new `DatadogLambda` class, so its behavior will be the same as the old `Datadog` class. Let me know if you think I should wrap it in a different way.
5. Create an example stack `lambda-function-stack-legacy-datadog-api.ts`, which uses `Datadog` instead of `DatadogLambda`, showing how to use the legacy `Datadog` API
6. Create a test `datadog.spec.ts` to ensure the `Datadog` API still works

### Motivation

To make it cleaner to add `DatadogStepFunction` class later. Details in [[RFC] Changing API for Datadog CDK Construct](https://docs.google.com/document/d/1g3bFuhWUayvuS_aS6kOHe_CGECocd8IhoKmM9s4dPU0)

#### Automated testing
1. pass existing automated tests
2. pass the new `datadog.spec.ts`
##### 1. using `DatadogLambda` API
Steps:
1. In `examples/typescript-stack/lib/cdk-typescript-stack.ts`, replace the `CdkTypeScriptStack` class with the `ExampleStack` class in `integration_tests/stacks/lambda-function-stack.ts`. I can't use the existing `CdkTypeScriptStack` class due to some docker permission issue.
    - The `ExampleStack` class uses `DatadogLambda` API
2. Follow the instructions in [Datadog CDK Constructs internal wiki](https://datadoghq.atlassian.net/wiki/spaces/SLS/pages/2723906703/Datadog+CDK+Constructs#Testing) to install the local package deploy the stack

Result:
- Deployment is successful. The Lambda functions have Datadog layers installed.

##### 2. using `Datadog` API
Step:
- In the test above, make the `ExampleStack` class use `Datadog` class instead of `DatadogLambda`. Deploy again.

Result:
- Deployment is also successful. The Lambda functions have Datadog layers installed.

### Additional Notes

Next steps:
1. update `README.md`, recommending users to use `DatadogLambda` instead of `Datadog`

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

[...]

PR #304, opened 2024-10-03
feat: Redo: Rename Lambda related interfaces

lym953:
### What does this PR do?
Redoes https://github.com/DataDog/datadog-cdk-constructs/pull/288:
1. Rename: `DatadogProps` -> `DatadogLambdaProps`
2. Rename: `DefaultDatadogProps` -> `DatadogLambdaDefaultProps`
3. Rename: `DatadogStrictProps` -> `DatadogLambdaStrictProps`
4. For backward compatibility, add:
```
export interface DatadogProps extends DatadogLambdaProps {}
export const DatadogDefaultProps = DatadogLambdaDefaultProps;
export interface DatadogStrictProps extends DatadogLambdaStrictProps {}
```
5. Add an example Python stack and an example Go stack that use the new APIs (`DatadogLambda` and `DatadogLambdaProps`) to `examples/`. Also add these two stacks to `integration_tests/stacks` for snapshot tests.

### Motivation
To make it cleaner to add `DatadogStepFunction` class later. Details in [[RFC] Changing API for Datadog CDK Construct](https://docs.google.com/document/d/1g3bFuhWUayvuS_aS6kOHe_CGECocd8IhoKmM9s4dPU0)

#### 1. Pass snapshot tests
`aws-vault exec sso-serverless-sandbox-account-admin -- scripts/run_integration_tests.sh`
The stacks can be processed without error.

#### 2. Notes on mixed API use
One use case will be broken by this PR: it's when the user uses Golang and uses `DatadogProps` (old) to construct `DatadogLambda` (new, added on Aug 29 in https://github.com/DataDog/datadog-cdk-constructs/pull/285), i.e.
```
	datadog := ddcdkconstruct.NewDatadogLambda(
		stack,
		jsii.String("Datadog"),
		&ddcdkconstruct.DatadogProps{
			...
		})
```
because `NewDatadogLambda()` can only take `DatadogLambdaProps`.

However, I think this case should be rare because:
1. The Go example code in `README.md` is still using the old `Datadog` instead of `DatadogLambda`
2. Go support was just added on Aug 23, so the user base should be small.

I'm also unable to find a good way to make this case happy. I'd just go ahead and release it, and ask users to update their code if the rare case does exists.

#### 3. Notes on other types/constants
I don't think people are supposed to use `DefaultDatadogProps` and `DatadogStrictProps` directly, so I didn't test the various combinations around those consts/interfaces. There's a chance that user code will be broken if `DefaultDatadogProps` or `DatadogStrictProps` are use directly. Let's closely monitor the issues after the release. In case anyone complains, let's fix forward (revert this PR then release), like what we did last time.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/examples/go-stack/app.go
+++ b/examples/go-stack/app.go
@@ -48,8 +48,38 @@ func NewAppStackWithoutDatadog(scope constructs.Construct, id *string, props *Ap
 	return stack, myFunction
 }
 
-// Creates a stack with Datadog integration set up
-func NewAppStackWithDatadog(scope constructs.Construct, id string, props *AppStackProps) awscdk.Stack {
+// Creates a stack with Datadog integration set up, using the new API (DatadogLambda, DatadogLambdaProps)
+func NewAppStackWithDatadogLambda(scope constructs.Construct, id string, props *AppStackProps) awscdk.Stack {
+	stack, lambdaFunction := NewAppStackWithoutDatadog(scope, &id, props)
+
+	// Set up Datadog integration
+	datadog := ddcdkconstruct.NewDatadogLambda(
+		stack,
+		jsii.String("Datadog"),
+		&ddcdkconstruct.DatadogLambdaProps{
+			NodeLayerVersion:      jsii.Number(113),
+			PythonLayerVersion:    jsii.Number(97),
+			JavaLayerVersion:      jsii.Number(21),
+			DotnetLayerVersion:    jsii.Number(15),
+			AddLayers:             jsii.Bool(true),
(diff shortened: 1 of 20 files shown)


purple4reina (changes requested):
We need some sort of automated testing for python and golang as well. The lack of such tests are what led to the incident a few weeks ago.

lym953:
@purple4reina Added snapshot tests for Go and Python in the last few PRs. This PR is now ready for review again.

purple4reina (approved):

PR #305, opened 2024-10-08
chore: Refactor run_integration_tests.sh by adding stack language check

lym953:
### Background
- `integration_tests/stacks/` contains a few example Typescript stacks
- `integration_tests/snapshots/` contains the expected snapshots generated from the stacks
- `scripts/run_integration_tests.sh` generates snapshots from the stacks, and compare the generated stacks with expected stacks

### Motivation
The test only covers TypeScript stacks. We want to cover Python and Go stacks as well to prevent issues like https://github.com/DataDog/datadog-cdk-constructs/issues/294, where the Python package was broken.

### What does this PR do?
1. Move the Typescript stacks to a `typescript/` directory
2. Make `run_integration_tests.sh` check the language of each stack, instead of assuming all stacks are in Typescript. Make the script require that path of a typescript stack should be like `typescript/<name>.ts`.

### Testing Guidelines
Run `aws-vault exec sso-serverless-sandbox-account-admin -- scripts/run_integration_tests.sh`. It finished successfully and the output was same as before.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/run_integration_tests.sh
+++ b/scripts/run_integration_tests.sh
@@ -8,10 +8,19 @@
 
 set -e
 
-# To add new tests create a new ts file in the 'integration_tests/stacks' directory, append its file name to the STACK_CONFIGS array.
+# To add new tests create a new ts file in the 'integration_tests/stacks' directory, append its file path to the STACK_CONFIG_PATHS array.
 # Note: Each ts file will have its respective snapshot built in the snapshots directory, e.g. lambda-function-stack.ts
 #       will generate both snapshots/test-lambda-function-stack-snapshot.json and snapshots/correct-lambda-function-stack-snapshot.json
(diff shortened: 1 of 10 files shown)


scripts/run_integration_tests.sh
```diff
@@ -8,10 +8,19 @@
...
+STACK_CONFIG_PATHS=(
+    "typescript/lambda-provided-stack.ts"
+    "typescript/lambda-provided-arm-stack.ts"
+    "typescript/lambda-singleton-function-stack.ts"
+    "typescript/lambda-function-arm-stack.ts"
+    "typescript/lambda-function-stack.ts"
+    "typescript/lambda-nodejs-function-stack.ts"
+    "typescript/lambda-python-function-stack.ts"
+    "typescript/lambda-java-function-stack.ts"
+)
```
purple4reina, line 23: thank you for this change, such an improvement

purple4reina (approved):

PR #307, opened 2024-10-09
chore: Add integration test for Python stack

lym953:
### What does this PR do?
Add a Python stack to integration tests, so that the test will compare the snapshots generated after vs before, to check if a PR causes unexpected changes on the stack to be deployed.

The stack was copied from `examples/python-stack`. I had to copy it because I was unable to find a way to just reference the existing stack.

### Motivation
Adding more tests to make it safer to deploy changes.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/integration_tests/snapshots/correct-lambda_python_stack-snapshot.json
+++ b/integration_tests/snapshots/correct-lambda_python_stack-snapshot.json
@@ -1,0 +1,692 @@
+{
+ "Resources": {
+  "hellonodeServiceRole01BB45E0": {
+   "Type": "AWS::IAM::Role",
+   "Properties": {
+    "AssumeRolePolicyDocument": {
+     "Statement": [
+      {
+       "Action": "sts:AssumeRole",
+       "Effect": "Allow",
+       "Principal": {
+        "Service": "lambda.amazonaws.com"
+       }
+      }
+     ],
+     "Version": "2012-10-17"
+    },
+    "ManagedPolicyArns": [
+     {
+      "Fn::Join": [
+       "",
+       [
+        "arn:",
+        {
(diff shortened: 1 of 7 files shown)


integration_tests/snapshots/correct-lambda_python_stack-snapshot.json
purple4reina, line 1: Is this file suppose to be empty?

purple4reina, line 1: Is this file suppose to be empty?

lym953: Good catch! Fixed.

lym953: Good catch! Fixed.

lym953: Done!

integration_tests/snapshots/test-lambda_python_stack-snapshot.json
purple4reina, line 1: Does it seem weird to you that these `integration_tests/snapshots/test-*` files are committed to git? Seems like they're overwritten by the test runner anyway.

purple4reina (approved):
There's still an integration test failure, but other than that this looks great!

integration_tests/snapshots/test-lambda_python_stack-snapshot.json
lym953: Good point. I'm already working on that. Coming soon in the next PR!

PR #321, opened 2024-11-07
feat: [Step Function] Set tags

lym953:
### What does this PR do?
Enables setting tags for the state machine, including DD tags and custom tags. For example, if the user passes these properties to Datadog:
```
    const datadogSfn = new DatadogStepFunctions(this, "Datadog", {
      env: "dev",
      service: "my-service",
      version: "1.0.0",
      forwarderArn: process.env.DD_FORWARDER_ARN,
      tags: "custom-tag-1:tag-value-1,custom-tag-2:tag-value-2",
    });
    datadogSfn.addStateMachines([stateMachine]);
```
Then the state machine will have these tags:
- `DD_ENV`: `env`
- `DD_SERVICE`: `my-service`
- `DD_VERSION`: `1.0.0`
- `custom-tag-1`: `tag-value-1`
- `custom-tag-2`: `tag-value-2`

### Motivation
This is one step of instrumenting a step function.

#### Automated testing
Passed the added test

#### Manual testing

Deployed the example stacks `step-functions-typescript-stack`. The tags are set as expected.
image

### Additional Notes
Need a stamp from Cloud Tracing team, who, as the downstream consumer, should know what tags should be set here.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/datadog-step-functions.ts
+++ b/src/datadog-step-functions.ts
@@ -14,6 +14,7 @@ import { Construct } from "constructs";
 import log from "loglevel";
 import { addForwarderForStateMachine } from "./forwarder";
 import { DatadogStepFunctionsProps } from "./index";
+import { setTags } from "./tag";
 
 const unsupportedCaseErrorMessage =
   "Step Function Instrumentation is not supported. \
@@ -45,6 +46,8 @@ export class DatadogStepFunctions extends Construct {
     } else {
       log.debug("Forwarder ARN not provided, no log group subscriptions will be added");
     }
+
+    setTags(stateMachine, this.props);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/datadog-step-functions.ts
```diff
@@ -155,6 +158,28 @@ export class DatadogStepFunctions extends Construct {
       )
     );
   }
+
+  private setTags(stateMachine: sfn.StateMachine): void {
```
lym953, line 162: Modified from https://github.com/DataDog/datadog-cdk-constructs/blob/main/src/datadog-lambda.ts#L181-L205. Not reusing it because the tag keys are different (e.g. `env` vs `DD_ENV`).

src/datadog-step-functions.ts
```diff
@@ -155,6 +158,28 @@ export class DatadogStepFunctions extends Construct {
...
+    const props = this.props;
+    if (props.env) {
+      Tags.of(stateMachine).add(DD_ENV_ENV_VAR, props.env);
+    }
+    if (props.service) {
+      Tags.of(stateMachine).add(DD_SERVICE_ENV_VAR, props.service);
+    }
+    if (props.version) {
+      Tags.of(stateMachine).add(DD_VERSION_ENV_VAR, props.version);
```
purple4reina, line 171: My understanding is that these three need to be added to the function as env vars, not as part of the dd tags. But I also don't have a good enough understanding of our step function instrumentation. Can you help me understand the decision to place all the tags in the location you did?

purple4reina (approved):

PR #398, opened 2025-04-01
Update interface

gabedos:
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/ecs/utils.ts
```diff
@@ -63,7 +63,11 @@ export function validateECSBaseProps(props: DatadogECSBaseProps): void {
   }
 }
 
-export function isOperatingSystemLinux(props: ecs.FargateTaskDefinitionProps): boolean {
+export function isOperatingSystemLinux(props: ecs.FargateTaskDefinitionProps | undefined): boolean {
```
avangelillo, line 66: Usually I find it's easier to handle defaulting the potentially undefined optional prop close to where it is first passed in, that way you don't need to handle it everywhere it ends up gets passed to

avangelillo (approved):

src/ecs/utils.ts
gabedos: Agreed! This method is only used in the constructor to get the "complete" default props.

PR #419, opened 2025-04-23
chore: Track agent install info [CONTP-776]

gabedos:
### What does this PR do?

Updates the Agent install tool information to accurately reflect the CDK construct and version.

### Motivation

Previously, the install info would default to the docker daemon that installs the container. 

### Testing Guidelines

N/A

### Additional Notes

N/A

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/ecs/fargate/environment.ts
+++ b/src/ecs/fargate/environment.ts
@@ -18,9 +18,9 @@ export class FargateEnvVarManager extends EnvVarManager {
 
     this.addAll(props.environmentVariables);
 
-    this.add("DD_INSTALL_INFO_TOOL", "datadog-cdk-constructs");
-    this.add("DD_INSTALL_INFO_TOOL_VERSION", versionJson.version);
-    this.add("DD_INSTALL_INFO_INSTALLER_VERSION", "");
+    this.add("DD_INSTALL_INFO_TOOL", "cdk");
+    this.add("DD_INSTALL_INFO_TOOL_VERSION", "datadog-cdk-constructs");
+    this.add("DD_INSTALL_INFO_INSTALLER_VERSION", versionJson.version);
 
     this.add("DD_API_KEY", props.apiKey);
     this.add("DD_SITE", props.site);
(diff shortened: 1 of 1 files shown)


src/ecs/fargate/environment.ts
```diff
@@ -17,6 +18,10 @@ export class FargateEnvVarManager extends EnvVarManager {
 
     this.addAll(props.environmentVariables);
 
+    this.add("DD_INSTALL_INFO_TOOL", "datadog-cdk-constructs");
+    this.add("DD_INSTALL_INFO_TOOL_VERSION", versionJson.version);
+    this.add("DD_INSTALL_INFO_INSTALLER_VERSION", "");
```
clamoriniere, line 23: I feel like the usage of the version is not aligned with other installer like [helm-chart](https://github.com/DataDog/helm-charts/blob/4509a10c2d94f561a1bb811dfe1efe5820f8863f/charts/datadog/templates/install_info-configmap.yaml#L14-L16) and the operator.
In the helm chart the `DD_INSTALL_INFO_TOOL_VERSION` is set with `.Release.Service` which can be either with the value `helm` for helmv3 and `tiller` for helmv2. I suppose other tools that can use a helm chart as source (fluxcd or argocd) might set different values.

from what I understand `DD_INSTALL_INFO_TOOL_VERSION` should be CDK version or flavour  and `DD_INSTALL_INFO_INSTALLER_VERSION` the version the construct

so maybe 

```
this.add("DD_INSTALL_INFO_TOOL", "cdk");
this.add("DD_INSTALL_INFO_TOOL_VERSION", "datadog-cdk-constructs" );
this.add("DD_INSTALL_INFO_INSTALLER_VERSION", versionJson.version);
```

if we want to be aligned with the helm-chart. It can be good to ask agent-onboarding team for recommendation.

gabedos: Agree we can align closer to what the helm / operator does.

```
% k get configmap datadog-agent-linux-installinfo -n datadog-agent -o yaml
apiVersion: v1
data:
  install_info: |
    ---
    install_method:
      tool: helm
      tool_version: Helm
      installer_version: datadog-3.110.13
```

The variable names are just a little confusing because I'd assume `tool_version` should have some version attached to it.

I tried to see if we could dynamically figure out what version of `cdk` the user generated the deployment with but that doesn't seem to be possible in a reliable manner.

adel121 (approved):

PR #425, opened 2025-05-05
fix the repository_url tag for gitlab.com

encron:
### What does this PR do?

The `git.repository_url` tag can not contain any protocols. For Github, this is being handled already but not for Gitlab (Saas). This PR applies the same normalisation for `gitlab.com`.

### Motivation

Received the following error message: 
Screenshot 2025-05-05 at 15 17 03

Noticed that the `repository_url` tag is being set to `https` and realized that it is setting the full URL (including the protocol) in the `DD_TAGS` variable of the Lambda function. 

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -76,17 +76,17 @@ function getGitData(): { hash: string; gitRepoUrl: string } {
     log.debug(`Failed to add source code integration. Error: ${e}`);
     return { hash: "", gitRepoUrl: "" };
   }
-  return { hash, gitRepoUrl: filterAndFormatGithubRemote(gitRepoUrl) };
+  return { hash, gitRepoUrl: filterAndFormatGitRemote(gitRepoUrl) };
 }
 
 // Removes sensitive info from the given git remote url and normalizes the url prefix.
 // "git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/"
-function filterAndFormatGithubRemote(rawRemote: string): string {
+function filterAndFormatGitRemote(rawRemote: string): string {
   rawRemote = filterSensitiveInfoFromRepository(rawRemote);
   if (!rawRemote) {
     return rawRemote;
   }
-  rawRemote = rawRemote.replace(/git@github\.com:|https:\/\/github\.com\//, "github.com/");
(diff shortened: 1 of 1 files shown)


purple4reina (changes requested):
Thank you so much for your contribution! Can you please add tests for your change? The `test/env.spec.ts` file should be the right place. Would just like to see some tests to ensure that we're still replacing these strings as expected for both `github.com` and `gitlab.com`.

src/env.ts
```diff
@@ -76,17 +76,17 @@ function getGitData(): { hash: string; gitRepoUrl: string } {
...
 // Removes sensitive info from the given git remote url and normalizes the url prefix.
 // "git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/"
-function filterAndFormatGithubRemote(rawRemote: string): string {
+function filterAndFormatGitRemote(rawRemote: string): string {
   rawRemote = filterSensitiveInfoFromRepository(rawRemote);
   if (!rawRemote) {
     return rawRemote;
   }
-  rawRemote = rawRemote.replace(/git@github\.com:|https:\/\/github\.com\//, "github.com/");
+  rawRemote = rawRemote.replace(/git@(github\.com|gitlab\.com):|https:\/\/(github\.com|gitlab\.com)\//, "$1$2/");
```
lym953, line 89: For clarity, could you separate the two cases, like:
```
rawRemote = rawRemote
  .replace(/git@(github\.com|gitlab\.com):/, "$1/")
  .replace(/https:\/\/(github\.com|gitlab\.com)\//, "$1/");
```

lym953 (commented):
Thanks for the change!

src/env.ts
encron: Done!

encron: I was actually wondering if there's a specific reason to only do this for `github.com` (and now also `gitlab.com`). In theory a git repo could be hosted anywhere and I could easily change this `replace` to just remove the protocol prefix instead, regardless of the host.

purple4reina (approved):

lym953 (approved):

PR #426, opened 2025-05-05
Instrumenting API Gateway inferred spans.

purple4reina:
### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/ecs/fargate/README.md
+++ b/src/ecs/fargate/README.md
@@ -172,10 +172,11 @@ For more general information, reference the [Datadog ECS Fargate Docs](https://d
 
 | Property                   | Type          | Description                                                                      |
 | -------------------------- | ------------- | -------------------------------------------------------------------------------- |
-| `isEnabled`                | `boolean`     | Enables DogStatsD.                                                               |
-| `isOriginDetectionEnabled` | `boolean`     | Enables DogStatsD origin detection.                                              |
-| `dogstatsdCardinality`     | `Cardinality` | Controls the cardinality of custom DogStatsD metrics.                            |
-| `isSocketEnabled`          | `boolean`     | Enables DogStatsD traffic over Unix Domain Socket. Falls back to UDP when false. |
+| `isEnabled`                  | `boolean`     | Enables DogStatsD.                                                               |
+| `isOriginDetectionEnabled`   | `boolean`     | Enables DogStatsD origin detection.                                              |
+| `dogstatsdCardinality`       | `Cardinality` | Controls the cardinality of custom DogStatsD metrics.                            |
+| `isSocketEnabled`            | `boolean`     | Enables DogStatsD traffic over Unix Domain Socket. Falls back to UDP when false. |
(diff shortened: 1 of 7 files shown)


src/ecs/fargate/datadog-ecs-fargate.ts
```diff
@@ -199,6 +199,14 @@ export class DatadogECSFargateTaskDefinition extends ecs.FargateTaskDefinition {
       container.addEnvironment("DD_AGENT_HOST", "127.0.0.1");
     }
 
+    // APM configuration
+    const traceInferredServices = this.datadogProps.apm!.traceInferredProxyServices;
+    if (this.datadogProps.apm!.isEnabled && traceInferredServices !== undefined) {
+      if (props.environment?.DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED === undefined) {
```
gabedos, line 205: Do we need this second if condition to prevent overriding of the environment value set by the user? Generally, the construct has precedence over any user defined values to guarantee their task is configured in the way defined via `datadogProps`.

The main use case I could imagine this is if users want to individually disable the feature on some containers and have it enabled on some other containers within the same task.

gabedos (approved):
Looks good to me.

Could add some tests to `test/ecs/fargate/apm-dsd.spec.ts` with the feature enabled and ensuring envvar is added to the application containers. We do similar for `DD_DOGSTATSD_URL`, `DD_TRACE_AGENT_URL`, and `DD_AGENT_HOST` already in the tests

src/ecs/fargate/datadog-ecs-fargate.ts
purple4reina: Alright, that's good feedback. I think it's probably best to stick with the precedent you already set.

gabedos (approved):
Looks good on the ECS Fargate construct. Thanks for updating unit tests and adding the first snapshot tests for ECS Fargate.

src/ecs/fargate/README.md
```diff
@@ -170,12 +170,13 @@ For more general information, reference the [Datadog ECS Fargate Docs](https://d
...
+| `isOriginDetectionEnabled`   | `boolean`     | Enables DogStatsD origin detection.                                                                                                                                                              |
+| `dogstatsdCardinality`       | `Cardinality` | Controls the cardinality of custom DogStatsD metrics.                                                                                                                                            |
+| `isSocketEnabled`            | `boolean`     | Enables DogStatsD traffic over Unix Domain Socket. Falls back to UDP when false.                                                                                                                 |
+| `traceInferredProxyServices` | `boolean`     | Enables inferred spans for proxy services like AWS API Gateway. When enabled, the tracer will create spans for proxy services by using headers passed from the proxy service to the application. |
```
clamoriniere, line 179: is it the right place for this config option? 
I might wrong, but it seems to be more related to `APMFeatureConfig` than `DogstatsdFeatureConfig`.

TalUsvyatsky (approved):

src/ecs/fargate/README.md
purple4reina: Ah, thank you, good catch!

Issue #596, opened 2026-05-07
DatadogLambda v3.12.0 crashes with aws-cdk-lib 2.253.0 — TypeError in setGitEnvironmentVariables


hari-euka:
## Description

`datadog-cdk-constructs-v2@3.12.0` crashes during `cdk synth` when used with `aws-cdk-lib@2.253.0`. The error occurs in `setGitEnvironmentVariables()` when `sourceCodeIntegration` is enabled (the default).

## Error

```
node_modules/datadog-cdk-constructs-v2/lib/env.js:65
  lambdas.forEach((lam) => {
TypeError: Cannot read properties of undefined (reading 'value')
    at node_modules/datadog-cdk-constructs-v2/lib/env.js:71:20
    at Array.forEach (<anonymous>)
    at setGitEnvironmentVariables (env.js:65:11)
    at DatadogLambda.addGitCommitMetadata (datadog-lambda.js:225:31)
    at DatadogLambda.addLambdaFunctions (datadog-lambda.js:178:14)
```

## Root Cause

In `env.js` lines 65-71, `setGitEnvironmentVariables()` does:

```js
lambdas.forEach((lam) => {
    if (lam.environment[exports.DD_TAGS] !== undefined) {
        lam.environment[exports.DD_TAGS].value += `,git.commit.sha:${hash}`;
    } else {
        lam.addEnvironment(exports.DD_TAGS, `git.commit.sha:${hash}`);
    }
    lam.environment[exports.DD_TAGS].value += `,git.repository_url:${gitRepoUrl}`;  // <-- CRASH
});
```

The `else` branch calls `lam.addEnvironment()`, then immediately accesses `lam.environment[DD_TAGS].value` on the next line. In `aws-cdk-lib@2.253.0`, the internal `environment` property structure appears to have changed — after `addEnvironment()`, `lam.environment[DD_TAGS]` doesn't have a `.value` property (or is structured differently), causing the TypeError.

This worked with `aws-cdk-lib@2.251.0` but breaks with `2.253.0`.

## Steps to Reproduce

1. Create a CDK stack with a Python Lambda function
2. Use `datadog-cdk-constructs-v2@3.12.0` with `aws-cdk-lib@2.253.0`
3. Run `cdk synth`

```typescript
import { DatadogLambda } from "datadog-cdk-constructs-v2";

const datadog = new DatadogLambda(this, "Datadog", {
    pythonLayerVersion: 91,
    extensionLayerVersion: 55,
[...]

PR #597, opened 2026-05-07
fix: prevent TypeError in setGitEnvironmentVariables with CDK 2.253+

hari-euka:
## Summary

Fixes #596 — `DatadogLambda.addLambdaFunctions()` crashes with `TypeError: Cannot read properties of undefined (reading 'value')` when used with `aws-cdk-lib >= 2.253.0`.

## Root Cause

In `src/env.ts`, `setGitEnvironmentVariables()` accesses the internal `lam.environment[DD_TAGS].value` property unconditionally on line 71, after the `else` branch calls `lam.addEnvironment()`:

```typescript
// Before (broken with CDK 2.253+)
lambdas.forEach((lam) => {
    if (lam.environment[DD_TAGS] !== undefined) {
        lam.environment[DD_TAGS].value += `,git.commit.sha:${hash}`;
    } else {
        lam.addEnvironment(DD_TAGS, `git.commit.sha:${hash}`);
    }
    // This line crashes — after addEnvironment(), the internal
    // .environment structure in CDK 2.253 doesn't expose .value
    lam.environment[DD_TAGS].value += `,git.repository_url:${gitRepoUrl}`;
});
```

In `aws-cdk-lib@2.253.0`, the internal `environment` property structure on Lambda functions changed. After calling `addEnvironment()`, `lam.environment[DD_TAGS]` may not have a `.value` property, causing the TypeError.

## Fix

Use `addEnvironment()` via the public API consistently instead of directly mutating the internal `.environment` property:

```typescript
// After (works with all CDK versions)
const gitTags = `git.commit.sha:${hash},git.repository_url:${gitRepoUrl}`;

lambdas.forEach((lam) => {
    const existingDdTags = lam.environment[DD_TAGS]?.value;
    if (existingDdTags) {
        lam.addEnvironment(DD_TAGS, `${existingDdTags},${gitTags}`);
    } else {
        lam.addEnvironment(DD_TAGS, gitTags);
    }
});
```

Key changes:
- Uses optional chaining (`?.value`) to safely read existing value
- Builds the full git tags string upfront (both `git.commit.sha` and `git.repository_url`)
- Uses `addEnvironment()` for all writes — never directly mutates `.value`
- Simpler code — no split between setting sha and repo url

## Tests

All 240 existing tests pass, including the 4 `DD_TAGS` tests that verify:
- `git.commit.sha` and `git.repository_url` are set by default
- Existing `DD_TAGS` values are preserved when adding git metadata
- Source code integration can be disabled

## Workaround

[...]

PR #598, opened 2026-05-08
fix: bugs on setGitEnvironmentVariables

fed-sv:
### What does this PR do?

Fixes a `TypeError: Cannot read properties of undefined (reading 'value')` crash in `setGitEnvironmentVariables` when `DD_TAGS` is not already set on the lambda function.

Also fixes a minor log message bug in the `gitRepoUrlOverride` debug statement, which was logging `hash` instead of `gitRepoUrl` as the "instead of" value.

### Motivation

The `else` branch of the `DD_TAGS` check calls `lam.addEnvironment(DD_TAGS, ...)` to set the sha, then immediately reads back `lam.environment[DD_TAGS].value` on the following line to append the repo URL. Because `lam.environment` is a private CDK field accessed via `any` cast, this read-back is not guaranteed to reflect the value just written by `addEnvironment` - in particular, it fails when there is a CDK peer dependency version mismatch between the consuming project and the version this library was tested against. The result is a hard crash at CDK synth time that completely prevents deployment.

The root cause is that the two git tags (`git.commit.sha` and `git.repository_url`) are set in two separate operations, but the second operation always uses the private field read - even when the first operation went through the public `addEnvironment` API (the `else` branch). 

### Testing Guidelines

Existing unit tests covering `setGitEnvironmentVariables` continue to pass. The fix is behaviorally identical for the case where `DD_TAGS` is already set (private field is read, confirmed non-undefined, then appended to - unchanged). For the case where `DD_TAGS` is not yet set, both git tags are now written atomically in a single `addEnvironment` call, eliminating the unsafe read-back.

### Additional Notes

The unconditional line 71 (`lam.environment[DD_TAGS].value += ...`) was always unsafe after the `else` branch because `addEnvironment` is a public method that may not synchronously update the private `environment` map in all CDK versions. Combining both tags into a single value in both branches removes this assumption entirely.

### Types of Changes

- [x] Bug fix

### Check all that apply

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -55,20 +55,20 @@ export function setGitEnvironmentVariables(
     hash = gitCommitShaOverride;
   }
   if (gitRepoUrlOverride) {
-    log.debug(`Using git repo URL override.  Will be ${gitRepoUrlOverride} instead of ${hash}`);
+    log.debug(`Using git repo URL override.  Will be ${gitRepoUrlOverride} instead of ${gitRepoUrl}`);
     gitRepoUrl = gitRepoUrlOverride;
   }
 
   if (hash == "" || gitRepoUrl == "") return;
 
   // We're using an any type here because AWS does not expose the `environment` field in their type
+  const gitTags = `git.commit.sha:${hash},git.repository_url:${gitRepoUrl}`;
   lambdas.forEach((lam) => {
-    if (lam.environment[DD_TAGS] !== undefined) {
-      lam.environment[DD_TAGS].value += `,git.commit.sha:${hash}`;
+    if (lam.environment[DD_TAGS] === undefined) {
+      lam.addEnvironment(DD_TAGS, gitTags);
     } else {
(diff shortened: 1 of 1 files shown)


hari-euka:
I raised a PR to address the exact same issue this morning @fed-sv 
https://github.com/DataDog/datadog-cdk-constructs/pull/597 😄

fed-sv:
> I raised a PR to address the exact same issue this morning @fed-sv #597 😄

ohhh LOL sorry I should have check that! happy for mine to be nuked then 💥. We really need this bug fixed asap 🥺
Thanks 🚀

fed-sv:
@hari-euka sorry I didn't realize your PR was closed 🤣

fed-sv:
@avangelillo  feel free to close this PR if you guys rather go with #597 👍

fed-sv:
add missing link to issue. Fixes #596

TalUsvyatsky:
Thanks for opening this PR, this was ultimately resolved with #601 and https://github.com/DataDog/datadog-cdk-constructs/releases/tag/v2-4.0.0

PR #621, opened 2026-06-23
fix: stop reading private aws-cdk-lib Function.environment field

### LINKED ISSUE #620: `datadog-cdk-constructs-v2` reads aws-cdk-lib's **private** `Function.environment` internals — it broke `cdk synth` once (2.253.0), and the "fix" dug deeper instead of respecting the interface
## Summary

This library reaches into `aws-cdk-lib`'s **private** `Function.environment` field — and, since v4.0.0, into the **private `.map` backing store** of the internal `Box` that field now holds. The code's own comments admit it: *"AWS does not expose the `environment` field"* and *"cast to any to access the private environment fields."* It does so anyway.

That is the whole bug. A routine, semver-compatible aws-cdk-lib minor (2.252 → 2.253) changed the field's internal shape and broke `cdk synth` for every stack using the default `sourceCodeIntegration` (#596). The v4.0.0 "fix" did not remove the illegal access — it re-pointed it at the *deeper* new internal and raised the peer floor, then closed the follow-on breakage as the user's fault (#606). It will break again on the next refactor.

## The defect (current `main`)

[`src/env.ts#L67`](https://github.com/DataDog/datadog-cdk-constructs/blob/61564e3c88efa5d68b0a37c86b0cb0d2c68a8fd9/src/env.ts#L67) and [`#L124`](https://github.com/DataDog/datadog-cdk-constructs/blob/61564e3c88efa5d68b0a37c86b0cb0d2c68a8fd9/src/env.ts#L124):

```ts
lam.environment.map.get(DD_TAGS)?.value          // setGitEnvironmentVariables
lam_with_env_vars.environment.map.get(envVar)    // applyEnvVariables
```

`environment` is and always has been `private`. aws-cdk-lib's public contract for env vars is `Function.addEnvironment(key, value)` — which this library already uses to *write*. The breakage exists only because it also *reads* across the private boundary.

| aws-cdk-lib | `Function.environment` |
|---|---|
| ≤ 2.252.0 | `private environment: { [key]: EnvironmentConfig } = {};` |
| ≥ 2.253.0 | `private readonly environment: IMapBox<…> = Box.fromMap(new Map());` |

(Verified at aws-cdk tags `v2.252.0` / `v2.253.0`.)

## The correct fix

1. **Respect the interface.** Stop reading the private field. This library is the sole writer of `DD_TAGS` and the `DD_*` defaults on these paths, so it can track its own writes (a `WeakMap<Function, …>`) instead of interrogating aws-cdk-lib's private state.
2. **Extend the interface, properly.** For the one case that benefits from reading user-set values, aws-cdk-lib has no public accessor *yet* — but it is open source and accepts contributions ([aws/aws-cdk](https://github.com/aws/aws-cdk)). Adding a supported `hasEnvironment(key)` / read-only getter is a small, routine, mergeable PR. The right path was always: collaborate upstream to add the public API, then depend on it. Reaching into `Box.map` instead — and, when that broke, reaching in *further* rather than opening that PR — reflects either a lack of diligence or a lack of competence, not a genuine constraint. There was a correct option available the entire time and it was not taken.

## Affected versions

- **datadog-cdk-constructs-v2**: ≤ 3.12.0 crashes on aws-cdk-lib ≥ 2.253.0; 4.0.0 / 4.1.0 keep the private-internal coupling and peer-lock to aws-cdk-lib ≥ 2.253.x.
- **aws-cdk-lib**: field representation changed in **2.253.0**; the construct is fragile on both sides of that line (#596, #606).

## Repro / impact

`sourceCodeIntegration` is on by default, so a transitive aws-cdk-lib bump to 2.253.0 turned `cdk synth` into a hard failure for downstream stacks (≤3.12.0):

```
TypeError: Cannot read properties of undefined (reading 'value')
    at setGitEnvironmentVariables (env.js)
    at DatadogLambda.addLambdaFunctions (datadog-lambda.js)
```

The only workarounds were disabling a default Datadog feature or pinning aws-cdk-lib. v4.0.0 converted this into a permanent peer constraint and a recurring fragility. Removing the dependency on private internals eliminates the entire class of failure.

This is also being raised with our Datadog account team. For a library published under the Datadog name that instruments production fleets by default, depending on a dependency's private internals — when a supported public path exists and a clean upstream extension was available — is below the standard we expect. We're asking for the root-cause fix, not another patch that re-points the same illegal access.

ava-silver:
### What does this PR do?

Removes all reads of `aws-cdk-lib`'s private `Function.environment` field from the library. Every read is replaced with a `WeakMap`-backed tracker that records the library's own `addEnvironment` calls. Adds a public `DatadogLambda.setEnvironment()` method so users can seed env vars the construct will respect.

### Motivation

Closes https://github.com/DataDog/datadog-cdk-constructs/issues/620

aws-cdk-lib changed `Function.environment`'s internal shape in a semver-minor (2.252 → 2.253), breaking `cdk synth` for all users of `sourceCodeIntegration` (#596). The v4.0.0 fix re-pointed the access at the new internal shape and raised the peer floor -- it kept the private coupling, so the library remained one CDK refactor away from breaking again.

Three call sites read across the private boundary:
- `env.ts` -- `setGitEnvironmentVariables` reads `DD_TAGS` to append git metadata
- `env.ts` -- `applyEnvVariables` checks each key before writing a default
- `datadog-lambda.ts` -- `overrideGitMetadata` reads `DD_TAGS` to rewrite git components

### Changes

**`src/env-tracker.ts` (new)**
- Module-level `WeakMap<LambdaFunction, Map<string, string>>` (`ddEnvTracker`) and three helpers -- `setTrackedEnv`, `getTrackedEnv`, `hasTrackedEnv` -- that record every env write the library makes. Internal, not re-exported from `index.ts`.

**`src/env.ts`**
- All `addEnvironment` calls in `setGitEnvironmentVariables`, `applyEnvVariables`, and `setDDEnvVariables` go through `setTrackedEnv`.
- Private field reads replaced with WeakMap lookups.
- `setGitEnvironmentVariables` parameter type tightened from `any[]` to `LambdaFunction[]`.

**`src/datadog-lambda.ts`**
- `overrideGitMetadata` reads `DD_TAGS` from `getTrackedEnv` instead of `lambdaFunction.environment.map.get(DD_TAGS)`, removing the `any` cast. Tag rewriting extracted to an `upsertTag` helper.
- Adds public `setEnvironment(lambdaFunction, key, value)`, which writes through `setTrackedEnv` so the construct treats the value as one it manages.

**`README.md`**
- Documents the env var ordering: configure `DD_*` vars via `DatadogLambdaProps`; call `func.addEnvironment()` *after* `addLambdaFunctions()` to override; or call `setEnvironment()` *before* to seed a value the construct will respect.

### Behavior change

Previously the library read the private field to detect whether the user had already set a `DD_*` var on a function before `addLambdaFunctions()`, skipping its default if so -- and to append git metadata onto a user-set `DD_TAGS`. Without a public CDK read API (which does not exist), arbitrary pre-set env vars are no longer visible to the construct. The new contract:

- **Configure via props** (recommended): use `DatadogLambdaProps` fields (`enableDatadogTracing`, `logLevel`, `tags`, etc.)
- **Seed a construct-respected value**: call `datadogLambda.setEnvironment(func, key, value)` *before* `addLambdaFunctions()`. The construct will not override it, and when source code integration is enabled it appends git metadata (`git.commit.sha`, `git.repository_url`) onto a seeded `DD_TAGS` rather than overriding it. This preserves the prior per-function `DD_TAGS` + git-metadata workflow.
- **Override after**: call `func.addEnvironment("DD_*", value)` *after* `addLambdaFunctions()` -- CDK's last-write-wins semantics handle this naturally.
- **Other vars set before `addLambdaFunctions()`** (not via `setEnvironment`): will be overridden by the library's defaults.

### Testing Guidelines

The existing suite covers the tracker migration; two tests were updated to reflect the behavior change (comments explain the new contract), and the `overrideGitMetadata` tests now assert via `Template.fromStack()` instead of reading the CDK private field. New tests cover `setEnvironment`: git metadata appending onto a seeded `DD_TAGS`, and the construct not overriding a seeded value.

Run locally with `yarn test`.

### Types of Changes

- [x] Bug fix
- [x] New feature
- [x] Breaking change

### Check all that apply

- [x] This PR contains breaking changes that are documented in the description
- [x] This PR impacts documentation, and it has been updated (or a ticket has been logged)

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ _Note_: The descriptions use the npm package parameters, but they also apply to 
 
 #### Setting `DD_*` environment variables
 
-To configure Datadog variables for every instrumented function, set the matching field on `DatadogLambdaProps` (for example, `enableDatadogTracing`, `logLevel`, `env`, `tags`).
+To configure Datadog variables for every instrumented function, set the matching field on `DatadogLambdaProps` (for example, `enableDatadogTracing`, `logLevel`, `env`, or `tags`).
 
 To override a value on a single function, use one of:
 
--- a/README.md
+++ b/README.md
@@ -404,6 +404,8 @@ The DatadogLambda construct takes in a list of lambda functions and installs the
 
 **Configuring `DD_*` environment variables:** Set Datadog environment variables through `DatadogLambdaProps` (e.g. `enableDatadogTracing`, `logLevel`, `tags`). If you need to set a `DD_*` variable directly on a function, call `func.addEnvironment()` **after** `addLambdaFunctions()` -- values set before will be overridden by the construct.
 
+To set a value the construct should respect (rather than override), use `datadogLambda.setEnvironment(func, key, value)` **before** `addLambdaFunctions()`. This is useful for setting `DD_TAGS` per function: when source code integration is enabled, the construct appends git metadata (`git.commit.sha`, `git.repository_url`) to the value you set instead of overriding it.
+
 While Lambda function based log groups are handled by the `addLambdaFunctions` method automatically, the construct has an additional function `addForwarderToNonLambdaLogGroups` which subscribes the forwarder to any additional log groups of your choosing.
 
 ## Step Functions
--- a/README.md
+++ b/README.md
@@ -287,13 +287,13 @@ To configure Datadog variables for every instrumented function, set the matching
 To override a value on a single function, use one of:
 
 - `datadogLambda.setEnvironment(func, key, value)` before `datadogLambda.addLambdaFunctions()`, to override a construct default while letting the construct finish instrumenting the function.
-- `func.addEnvironment(key, value)` after `datadogLambda.addLambdaFunctions()`, to set the final value and bypass the construct entirely.
+- `func.addEnvironment(key, value)` after `datadogLambda.addLambdaFunctions()`, to override the value set during instrumentation.
 
 When more than one source sets the same key, the following order applies (highest precedence first):
 
-1. `func.addEnvironment(key, value)` called after `datadogLambda.addLambdaFunctions()`
-2. `DatadogLambdaProps` fields dedicated to that key. These props always overwrite the same key set through. `datadogLambda.setEnvironment`:
-   - Unified service tagging: `env`, `service`, `version`, `tags`
+1. `func.addEnvironment(key, value)` called after `datadogLambda.addLambdaFunctions()`.
+2. `DatadogLambdaProps` fields dedicated to that key. These fields overwrite a value for the same key set through `datadogLambda.setEnvironment()`:
+   - Unified service tagging: `env`, `service`, `version`
    - Cold-start tracing: `enableColdStartTracing`, `minColdStartTraceDuration`, `coldStartTraceSkipLibs`
    - Other tracer settings: `enableProfiling`, `encodeAuthorizerContext`, `decodeAuthorizerContext`, `apmFlushDeadline`
    - LLM Observability: `llmObsEnabled`, `llmObsMlApp`, `llmObsAgentlessEnabled`
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ When `addLambdaFunctions` is called, the Datadog CDK construct grants your Lambd
 
 The DatadogLambda construct takes in a list of lambda functions and installs the Datadog Lambda Library by attaching the Lambda Layers for [.NET][19], [Java][15], [Node.js][2], and [Python][1] to your functions. It redirects to a replacement handler that initializes the Lambda Library without any required code changes. Additional configurations added to the Datadog CDK construct will also translate into their respective environment variables under each lambda function (if applicable / required).
 
-**Configuring `DD_*` environment variables:** Set Datadog environment variables through `DatadogLambdaProps` (e.g. `enableDatadogTracing`, `logLevel`, `tags`). If you need to set a `DD_*` variable directly on a function, call `func.addEnvironment()` **after** `addLambdaFunctions()` — values set before will be overridden by the construct.
+**Configuring `DD_*` environment variables:** Set Datadog environment variables through `DatadogLambdaProps` (e.g. `enableDatadogTracing`, `logLevel`, `tags`). If you need to set a `DD_*` variable directly on a function, call `func.addEnvironment()` **after** `addLambdaFunctions()` -- values set before will be overridden by the construct.
 
 While Lambda function based log groups are handled by the `addLambdaFunctions` method automatically, the construct has an additional function `addForwarderToNonLambdaLogGroups` which subscribes the forwarder to any additional log groups of your choosing.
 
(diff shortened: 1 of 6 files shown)


ava-silver:
* **#621** <a href="https://app.graphite.com/github/pr/DataDog/datadog-cdk-constructs/621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-cdk-constructs/621?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

README.md
```diff
@@ -409,6 +409,8 @@ When `addLambdaFunctions` is called, the Datadog CDK construct grants your Lambd
 
 The DatadogLambda construct takes in a list of lambda functions and installs the Datadog Lambda Library by attaching the Lambda Layers for [.NET][19], [Java][15], [Node.js][2], and [Python][1] to your functions. It redirects to a replacement handler that initializes the Lambda Library without any required code changes. Additional configurations added to the Datadog CDK construct will also translate into their respective environment variables under each lambda function (if applicable / required).
 
+**Configuring `DD_*` environment variables:** Set Datadog environment variables through `DatadogLambdaProps` (e.g. `enableDatadogTracing`, `logLevel`, `tags`). If you need to set a `DD_*` variable directly on a function, call `func.addEnvironment()` **after** `addLambdaFunctions()` -- values set before will be overridden by the construct.
```
janine-c, line 412:
```suggestion
To configure `DD_*` environment variables, set Datadog environment variables through `DatadogLambdaProps` (for example, `enableDatadogTracing`, `logLevel`, or `tags`). If you need to set a `DD_*` variable directly on a function, call `func.addEnvironment()` **after** `addLambdaFunctions()`. This construct overwrites any values that had been set before.
```

janine-c (approved):
Just some optional writing suggestions 🙂

ahammond:
Can we please get this merged?

ava-silver:
> Can we please get this merged?

We'll try to prioritize it, my main concern is that this version of the fix has a behavior change which may break some usage patterns. I have another potential option up as a separate PR, and I'll weight the options with the team before merging either.

Sorry for the delay, and thanks for your patience on this! I just want to make sure we're being careful before potentially releasing a breaking change.

ahammond:
Appreciate the caution, and I get that the behaviour change needs weighing. But ISTM this is being triaged as a minor cleanup. From the consumer side it is major.

The un-patched library reads a private aws-cdk-lib field, so every consumer of sourceCodeIntegration breaks whenever CDK reshapes that internal. #596 was triggered by a semver-minor (2.252 -> 2.253), and the v4.0.0 fix kept the private coupling, so the next break is a when, not an if. We're an engineering org with compliance and audit obligations: dependency upkeep is mandatory, automated, and continuous. A library that can break `cdk synth` fleet-wide on any CDK minor is a landmine sitting in that pipeline and has already cost me audit findings.

This PR has been open since June 23, and I want to be direct about the consequences. Next week I'll remove our usage of this library and with it Datadog observability support for lambdas. Future observability work will land in our in-house observability platform. I'd rather not deal with this right now, but Datadog's failure to address our ongoing exposure from this issue in a timely way is forcing my hand.

ava-silver:
@ahammond I've chatted with the team, and we'll go through with the current implementation, which includes a breaking change around setting environment variables. I'll work on getting this merged and released in the next couple days so you can migrate to the safer version.

avangelillo (approved):

TalUsvyatsky (approved):

ava-silver:
@ahammond we're working on some additional changes before we release the next version (with a major version bump). it should be out sometime next week.

PR #623, opened 2026-06-26
feat: default Lambda layer versions when none provided

### LINKED ISSUE #423: Provide constant for layer versions
## Expected Behavior

I would like to be able to set the `nodeLayerVersion` to a constant value provided by this library so that I don't have to update this version on a constant basis. The Datadog serverless plugin seems to handle this with a map file:

https://github.com/DataDog/serverless-plugin-datadog/blob/main/src/layers.json

#### Example
```typescript
const datadogLambda = new DatadogLambda(this, 'DatadogLambda', {
      flushMetricsToLogs: true,
      addLayers: true,
      nodeLayerVersion: NodeLayerVersion.Commerical.Latest, // Ideally a constant for the latest layer version is available. Perhaps the region must also be taken into account
    });
```

## Actual Behavior

```typescript
const datadogLambda = new DatadogLambda(this, 'DatadogLambda', {
      flushMetricsToLogs: true,
      addLayers: true,
      nodeLayerVersion: 124 // the version is hardcoded and must be looked up from https://github.com/DataDog/datadog-lambda-js/releases
    });
```

ava-silver:
## What

Closes #423.

`*LayerVersion` props are now optional. When `addLayers` is `true` (default) and neither a `*LayerVersion` nor `*LayerArn` is provided for a runtime, the construct applies a default layer version bundled with the package instead of erroring.

The defaults live in a new autogenerated `src/layer-versions.ts` and are exported so consumers can reference or pin them:

```ts
import { DD_DEFAULT_NODE_LAYER_VERSION } from "datadog-cdk-constructs-v2";
```

Available: `DD_DEFAULT_NODE_LAYER_VERSION`, `DD_DEFAULT_PYTHON_LAYER_VERSION`, `DD_DEFAULT_JAVA_LAYER_VERSION`, `DD_DEFAULT_DOTNET_LAYER_VERSION`, `DD_DEFAULT_RUBY_LAYER_VERSION`, `DD_DEFAULT_EXTENSION_LAYER_VERSION`.

These defaults track the latest released Datadog Lambda layers and are regenerated automatically.

## Behavior notes

- Passing a `*LayerVersion` or `*LayerArn` overrides the default (unchanged).
- The extension layer is unaffected -- it is still applied only when `extensionLayerVersion` or `extensionLayerArn` is set.
- `addLayers: false` still skips tracing layers entirely.
- The now-unreachable `getMissingLayerVersionErrorMsg`/`handleLayerError` helpers were removed.

## Test

- Rewrote the "missing version" test to assert the default layer ARN is attached and no errors are returned.
- Full suite passes (240 tests); `yarn eslint` and `yarn compile` clean.


ava-silver:
* **#623** <a href="https://app.graphite.com/github/pr/DataDog/datadog-cdk-constructs/623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-cdk-constructs/623?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

README.md
```diff
@@ -243,14 +233,14 @@ _Note_: The descriptions use the npm package parameters, but they also apply to
...
+| `pythonLayerVersion`         | `python_layer_version`          | Version of the Python Lambda layer to install, such as `122`. Optional: defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.PYTHON`). Set this to pin a specific version. Find the latest version number [here][5]. **Warning**: This parameter and `pythonLayerArn` are mutually exclusive. If used, only set one or the other.                                                                                                                                                                                                              |
```
jeff-morgan-dd:
```suggestion
| `pythonLayerVersion`         | `python_layer_version`          | Version of the Python Lambda layer to install, such as `122`. Optional: Defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.PYTHON`). Set this to pin a specific version. Find the latest version number [here][5]. **Warning**: This parameter and `pythonLayerArn` are mutually exclusive. If used, only set one or the other.                                                                                                                                                                                                              |
```

README.md
```diff
@@ -243,14 +233,14 @@ _Note_: The descriptions use the npm package parameters, but they also apply to
...
+| `nodeLayerVersion`           | `node_layer_version`            | Version of the Node.js Lambda layer to install, such as `133`. Optional: defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.NODE`). Set this to pin a specific version. Find the latest version number from [here][6]. **Warning**: This parameter and `nodeLayerArn` are mutually exclusive. If used, only set one or the other.                                                                                                                                                                                                            |
```
jeff-morgan-dd:
```suggestion
| `nodeLayerVersion`           | `node_layer_version`            | Version of the Node.js Lambda layer to install, such as `133`. Optional: Defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.NODE`). Set this to pin a specific version. Find the latest version number from [here][6]. **Warning**: This parameter and `nodeLayerArn` are mutually exclusive. If used, only set one or the other.                                                                                                                                                                                                            |
```

README.md
```diff
@@ -243,14 +233,14 @@ _Note_: The descriptions use the npm package parameters, but they also apply to
...
-| `javaLayerVersion`           | `java_layer_version`            | Version of the Java layer to install, such as `25`. Required if you are deploying at least one Lambda function written in Java and `addLayers` is `true`. Find the latest version number [here][28]. **Note**: `extensionLayerVersion >= 25` and `javaLayerVersion >= 5` are required for the DatadogLambda construct to instrument your Java functions properly. **Warning**: This parameter and `javaLayerArn` are mutually exclusive. If used, only set one or the other.                                                       |
+| `javaLayerVersion`           | `java_layer_version`            | Version of the Java layer to install, such as `25`. Optional: defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.JAVA`). Set this to pin a specific version. Find the latest version number [here][28]. **Note**: `extensionLayerVersion >= 25` and `javaLayerVersion >= 5` are required for the DatadogLambda construct to instrument your Java functions properly. **Warning**: This parameter and `javaLayerArn` are mutually exclusive. If used, only set one or the other.                |
```
jeff-morgan-dd:
```suggestion
| `javaLayerVersion`           | `java_layer_version`            | Version of the Java layer to install, such as `25`. Optional: Defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.JAVA`). Set this to pin a specific version. Find the latest version number [here][28]. **Note**: `extensionLayerVersion >= 25` and `javaLayerVersion >= 5` are required for the DatadogLambda construct to instrument your Java functions properly. **Warning**: This parameter and `javaLayerArn` are mutually exclusive. If used, only set one or the other.                |
```

README.md
```diff
@@ -243,14 +233,14 @@ _Note_: The descriptions use the npm package parameters, but they also apply to
...
+| `dotnetLayerVersion`         | `dotnet_layer_version`          | Version of the .NET layer to install, such as `23`. Optional: defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.DOTNET`). Set this to pin a specific version. Find the latest version number from [here][18]. **Warning**: This parameter and `dotnetLayerArn` are mutually exclusive. If used, only set one or the other.                                                                                                                                                                                                                  |
```
jeff-morgan-dd:
```suggestion
| `dotnetLayerVersion`         | `dotnet_layer_version`          | Version of the .NET layer to install, such as `23`. Optional: Defaults to the latest version bundled with this construct (`DatadogDefaultLayerVersions.DOTNET`). Set this to pin a specific version. Find the latest version number from [here][18]. **Warning**: This parameter and `dotnetLayerArn` are mutually exclusive. If used, only set one or the other.                                                                                                                                                                                                                  |
```

src/layer-versions.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/*
+ * Default Datadog Lambda layer versions bundled with this construct. These are
+ * used when a corresponding `*LayerVersion`/`*LayerArn` prop is not provided.
+ * Exposed as a class so the values are available in all jsii languages.
+ *
+ * DO NOT EDIT: This file is autogenerated.
+ */
+
+export class DatadogDefaultLayerVersions {
```
avangelillo, line 17: Because we already have the language passed into the `tryToFigureOutTracingLayerArn` function, if you make this a map of language -> version you don't need to add the other parameter to that function and can just look it up there

jeff-morgan-dd (approved):
Thank you! Please see a couple of nit style guide edits (all similar), but nothing blocking.

src/layer.ts
```diff
@@ -224,27 +225,29 @@ function tryToFigureOutTracingLayerArn(
   accountId: string | undefined,
   runtime: string,
   isARM: boolean,
-  lam: lambda.Function,
   errors: string[],
   formalRuntime: string,
   paramRuntime: string,
   layerVersion?: number,
   layerArn?: string,
+  defaultLayerVersion?: number,
```
avangelillo: Just to make sure explicitly, this isn't going to change behavior of anyone currently using this because we used to required a specific number and now we don't, previously we would have errored in this case?

ava-silver: correct! -- this is why i left out the extension version defaulting, since it would change behavior rather than error -> safe default

avangelillo (approved):

src/layer-versions.ts
ava-silver: done!

PR #646, opened 2026-07-27
chore: upgrade projen to 0.101.20 and opt back in to the system shell

lym953:
## What and why

Fixes the weekly `upgrade` workflow, [failing](https://github.com/DataDog/datadog-cdk-constructs/actions/runs/30227453075/job/89859951668) since projen 0.101.20 shipped with `dax: node: command not found`.

The task bumps `projen` (`^0.99.52` → `^0.101.20`) then re-runs `yarn projen`. projen 0.101.0 made its built-in shell the default for task commands and `$(...)` env vars ([deliberate break](https://github.com/projen/projen/pull/4770)), and the projen-generated `PATH: $(yarn exec node --print process.env.PATH)` can't resolve `yarn`/`node` under it.

The fix is the documented opt-out — `project.tasks.shell = TaskShell.system()` (see [projen/projen#4783](https://github.com/projen/projen/issues/4783)).

> Note: #648 succeeded only by luck — the `.ncurc.cjs` age gate skipped `projen` because 0.101.23 was ~2 hours short of 2 days old at the time. The break is still live for the next run.

## Silently-ignored options found along the way

Four options in `.projenrc.js` had been removed from projen's API and silently ignored (plain JS = no type error), so they were quietly doing nothing:

| Old option | Replacement | What had broken |
|---|---|---|
| `npmignore` | `npmIgnoreOptions.ignorePatterns` | all 12 custom `.npmignore` entries vanished |
| `scripts` | `project.addTask(...)` | `check-formatting` task vanished (CI uses it) |
| `releaseWorkflow: false` | `release: false` | `release.yml` + `bump`/`unbump` tasks appeared; `upgrade.yml` renamed |
| `mergify: false` | `githubOptions.mergify` | now defaults to `true` |

`jsiiFqn`, `releaseEveryCommit`, `rebuildBot` are gone with no replacement — dropped.

## Other forced changes

- `.ncurc.cjs` (2nd commit): it read the ncu package list from the upgrade task's `steps[0].exec`, which 0.101 renders as `execArgs`. The regex matched nothing, so the reject list came back empty and the 2-day supply-chain age gate was silently disabled. Now handles both shapes — verified it extracts all 16 filtered packages.
- `.projenrc.js` needs `eslint.allowDefaultProjectFiles` under 0.101's new `projectService` config.
- `tsconfig.dev.json` → `tsconfig.json` + `test/tsconfig.json` is projen's new generated layout (no opt-out); 0.101 also auto-assigns `id` to generated workflow steps.

## Verification

Locally: `yarn projen upgrade` exits 0 (the step that fails on `main`); `yarn projen` idempotent so anti-tamper passes; 240 tests + eslint + formatting clean; `compile` and `package:js`/`python`/`go` pass. `package:java` untested — no Maven locally; the `integration-tests` job covers it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.projenrc.js
```diff
@@ -70,29 +68,29 @@ const project = new awscdk.AwsCdkConstructLibrary({
     "*.tar.gz",
     "cdk.context.json",
   ],
-  npmignore: [
```
ava-silver, line 73: 💭 maybe we should switch this to be a .ts file such that we catch these errors sooner

ava-silver (approved):

PR #661, opened 2026-08-05
Do Not Merge, trying to debug the reason for projen failure

mrgrain:
This failure and change here should not be needed, trying to see what's going on:
https://github.com/DataDog/datadog-cdk-constructs/pull/646


mrgrain:
Found it: https://github.com/projen/projen/issues/4846#issuecomment-5192772213

